PR #8, opened 2024-09-17
Implement sunroof

WebSpider:
- Implement sunroof sensor

### DIFF
--- a/custom_components/myskoda/binary_sensor.py
+++ b/custom_components/myskoda/binary_sensor.py
@@ -251,6 +251,11 @@ class SunroofOpen(StatusBinarySensor):
 
     @property
     def is_on(self):  # noqa: D102
+        if (
+            self._status().detail.sunroof is None
+            or self._status().detail.sunroof == OpenState.UNSUPPORTED
+        ):
+            return None
         return self._status().detail.sunroof == OpenState.OPEN
 
     def is_supported(self) -> bool:
--- a/custom_components/myskoda/binary_sensor.py
+++ b/custom_components/myskoda/binary_sensor.py
@@ -42,6 +42,7 @@ async def async_setup_entry(
             LightsOn,
             ChargerConnected,
             ChargerLocked,
+            SunroofOpen,
         ],
         coordinator=hass.data[DOMAIN][config.entry_id][COORDINATOR],
         async_add_entities=async_add_entities,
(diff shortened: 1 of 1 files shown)


custom_components/myskoda/binary_sensor.py
```diff
@@ -317,6 +317,32 @@ def is_on(self):  # noqa: D102
...
+        self._attr_unique_id = f"{vehicle.info.vin}_sunroof_open"
+
+    @property
+    def is_on(self):  # noqa: D102
+        if not self.coordinator.data:
+            return None
+
+        self._update_device_from_coordinator()
+
+        return self.vehicle.status.detail.sunroof == OpenState.OPEN
```
Prior99, line 343: This way, when a vehicle doesn't have a sunroof, it will be reported as "CLOSED". I'd suggest returning `None` in that case for now.

We should find a way to leverage capabilities in the future, in order to not enable sensors that don't make sense.

Prior99 (changes requested):

custom_components/myskoda/binary_sensor.py
WebSpider: I'll create an issue for that so we dont forget, and patch this accordingly for now

custom_components/myskoda/binary_sensor.py
```diff
@@ -325,6 +325,35 @@ def is_on(self):  # noqa: D102
         return self.vehicle.status.detail.bonnet == OpenState.OPEN
 
 
+class SunroofOpen(MySkodaBinarySensor):
+    """Detectes whether the sunroof is open."""
+
+    def __init__(self, coordinator: DataUpdateCoordinator, vehicle: Vehicle) -> None:  # noqa: D107
```
dvx76, line 331: nit/later: if we're going to disable D107 for all/most init methods we might as well disable it globally in pyproject.toml

```
[tool.ruff.lint]
ignore = [
    "D171"   # undocumented-public-init
]
```

dvx76 (commented):
LGTM!

custom_components/myskoda/binary_sensor.py
Prior99: Agree

Prior99:
I rebased this and adapted it to the new structure.

Prior99 (approved):

PR #6, opened 2024-09-16
Add hassfest (prep for adding translations)

WebSpider:

PR #10, opened 2024-09-17
Translate config flow

WebSpider:
Translates basic config flow
Needs #6 so we can lint it from the start


Prior99:
Can you explain to me what this does?

Prior99 (approved):

PR #11, opened 2024-09-17
Fix unavailable device tracker

WebSpider:
When a car does not transmit it's coordinates, or has coordinates turned off, this causes an error in the current integration. Should be fixed with this update.

### DIFF
--- a/custom_components/myskoda/device_tracker.py
+++ b/custom_components/myskoda/device_tracker.py
@@ -8,7 +8,7 @@ from homeassistant.helpers.entity import EntityDescription
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
 from homeassistant.helpers.typing import DiscoveryInfoType
 from myskoda.models.info import CapabilityId
-from myskoda.models.position import Positions
+from myskoda.models.position import Position, PositionType, Positions
 
 from .const import COORDINATOR, DOMAIN
 from .coordinator import MySkodaDataUpdateCoordinator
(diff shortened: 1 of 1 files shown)


Prior99:
There is unrelated changes in the `poetry.lock` file.
I would suggest to search for a position with `type == PositionType.VEHICLE` instead of accessing the first element.

custom_components/myskoda/device_tracker.py
```diff
@@ -57,7 +57,10 @@ def latitude(self) -> float | None:  # noqa: D102
 
         self._update_device_from_coordinator()
 
-        return self.vehicle.position.positions[0].gps_coordinates.latitude
+        if len(self.vehicle.position.positions) > 0:
```
dvx76, line 60: nit: an non-empty list is True so you can just

```suggestion
        if self.vehicle.position.positions:
```

custom_components/myskoda/device_tracker.py
```diff
@@ -57,7 +57,10 @@ def latitude(self) -> float | None:  # noqa: D102
...
+            return self.vehicle.position.positions[0].gps_coordinates.latitude
+        else:
+            return None
```
dvx76, line 63: nit: if we enable all ruff rules (not done here yet) it will trip over https://docs.astral.sh/ruff/rules/superfluous-else-return/

```suggestion
        return None
```

dvx76 (approved):
LGTM with optional nits to address if you want.

Pet peeve: when updating from main do a rebase intead of merging main into the branch 😅

Prior99:
I rebased this and added a check to whether the position is of type `PositionType.VEHICLE`.

Prior99 (approved):

PR #19, opened 2024-09-19
Add precommit

WebSpider:
Just a basic pre-commit setup to extend upon

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+---
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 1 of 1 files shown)


dvx76 (approved):
Nice! LGTM although I guess the poetry.lock changes are unrelated. Maybe we don't want the `poetry-lock` hook to avoid exactly that?

WebSpider:
Well, we test in CI if the lockfile needs updating. The hook just makes sure you dont forget it.

Ive update the lockfile manually so many times to make CI succeed, i thought i'd add it ;)

Prior99 (approved):

PR #21, opened 2024-09-20
Add debouncing and switch to per-device coordinators

WebSpider:
Possibly this needs to be expanded into a debouncer per type of API call, unsure yet.


custom_components/myskoda/coordinator.py
```diff
@@ -1,29 +1,51 @@
...
+        super().__init__(
+            hass,
+            _LOGGER,
+            cooldown=API_COOLDOWN_IN_SECONDS,
+            immediate=False,
+            function=func,
+        )
+
 
 class State:
```
joostlek, line 43: Make this a dataclass, saves you a constrcutor

custom_components/myskoda/coordinator.py
```diff
@@ -36,63 +58,93 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
...
     data: State
-
-    def __init__(self, hass: HomeAssistant, config: ConfigEntry) -> None:
+    update_driving_range: RefreshFunction
+    update_charging: RefreshFunction
+    update_air_conditioning: RefreshFunction
+    update_vehicle: RefreshFunction
+    vin: str
+    hass: HomeAssistant
+
```
joostlek, line 67: Only define the types on class level if we can't infer the types

joostlek (changes requested):

custom_components/myskoda/coordinator.py
Prior99: Implemented in 3cac6d4

custom_components/myskoda/coordinator.py
Prior99: Implemented in 3cac6d4 except for `data: State` as discussed in discord

Prior99 (approved):

PR #40, opened 2024-09-27
Fix EN tranlations

WebSpider:
Stop using HA default referrals. Apparently they don't work for custom integrations
Fixes #30

PR #41, opened 2024-09-27
Fix door lockstate

### LINKED ISSUE #37: [Bug] lock status incorrect
With the more recent version the status of the "Doors locks" and "Locks" sensor is inverted.

My doors are locked as also shown in the app
image

In Home Assistant it shows:
myskoda_ha_locks

Oh, and maybe the sensor should be called "Door locks"

WebSpider:
According to https://www.home-assistant.io/integrations/binary_sensor/#device-class it's inverse from what you'd expect..
Fixes #37


WebSpider:
Merge it after #40 and everything should be OK.

dvx76 (approved):

PR #55, opened 2024-09-29
Add throttle to prevent API credit exhaustion

WebSpider:
Currently, every climate change, including a slider from the UI, results in a request to the car.
Given that the car has limited capabilities to handle many requests, we should limit those.

This introduces a HA mechanism called 'Throttle'. Basically, this means all requests after the first one are ignored for a timeperiod. Introducing this with 30s timeperiod.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,3 +52,14 @@ Air conditioning is exposed as climate.
 ### Device Tracker
 
 Location of vehicles are exposed as device trackers.
+
+## Sending updates
+
+To make sure updates end up in the car the way intended by the you, we throttle all equal changes to the car for 30s.
+Simply put: You can change a setting in the car once every 30s, but you can change multiple settings in sequence.
+
+So:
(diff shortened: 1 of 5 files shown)


dvx76:
2 comments:

1. Would we not want this on all outbound requests (e.g. `async_set_temperature`)
2. I think I would have preferred something which only throttles function calls when there is already a previous call executing. Which then basically forces the user to wait for the previous call to complete (and potentially result in an updated switch or sensor state). But if that doesn't exist and this is provided by ha-core then this is definitely the next best thing.

Base on the doc for `Throttle` (_Calling a method a second time during the interval will return None._) I guess the user experience is that just nothing happens. I think maybe we should document this (in the README) as being expected behavior?

dvx76 (approved):

WebSpider:
Good points. 
I'll add the entry to README and see if we have other calls that would benefit as well, besides the two we already identified.

dvx76 (approved):

PR #81, opened 2024-10-09
Add check to see if parking positions are not in-car disabled

### LINKED ISSUE #73: InvalidCapabilityConfigurationError when disabling position sharing in-car
**Describe the bug**
When the in-car setting is to not share positions, the integration logs an error in v1.4.5

Yup

**Latest working release**
N/A

**Debug logs**
```
2024-10-07 14:38:48.980 ERROR (MainThread) [custom_components.myskoda.coordinator] Unexpected exception from <bound method MySkodaDataUpdateCoordinator._update_vehicle of <custom_components.myskoda.coordinator.MySkodaDataUpdateCoordinator object at 0x795ca4d93080>>
  File "/home/me/dev/core/config/custom_components/myskoda/coordinator.py", line 183, in _update_vehicle
  File "/home/me/dev/core/config/custom_components/myskoda/coordinator.py", line 157, in set_updated_vehicle
  File "/home/me/dev/core/config/custom_components/myskoda/device_tracker.py", line 71, in latitude
  File "/home/me/dev/core/config/custom_components/myskoda/device_tracker.py", line 56, in _vehicle_position
  File "/home/me/dev/core/config/custom_components/myskoda/device_tracker.py", line 50, in _positions
custom_components.myskoda.utils.InvalidCapabilityConfigurationError: Entity 'VIN_device_tracker' has a bad capability configuration for vehicle 'Enyaq Engine(type='iV', power=150, capacity_in_liters=None) 2021 (5AZJJ2)'.
```

**Installation:**
 - Home Assistant variant: Core
 - Python version (when running Core or Supervised) 3.12

**Additional context**
Works fine when position sharing is enabled in the car

WebSpider:
Fixes #73

### DIFF
--- a/custom_components/myskoda/device_tracker.py
+++ b/custom_components/myskoda/device_tracker.py
@@ -53,7 +53,7 @@ class DeviceTracker(MySkodaEntity, TrackerEntity):
         return positions
 
     def _vehicle_position(self) -> Position | None:
-        if self._positions().positions:
+        if self._positions is not None and self._positions().positions:
             return next(
                 pos
                 for pos in self._positions().positions
--- a/custom_components/myskoda/device_tracker.py
+++ b/custom_components/myskoda/device_tracker.py
@@ -53,14 +53,13 @@ class DeviceTracker(MySkodaEntity, TrackerEntity):
         return positions
 
     def _vehicle_position(self) -> Position | None:
-        if self._positions is not None and self._positions().positions:  # pyright: ignore reportOptionalMemberAccess
-            return next(
-                pos
-                for pos in self._positions().positions  # pyright: ignore reportOptionalMemberAccess
-                if pos.type == PositionType.VEHICLE
-            )
-        else:
-            return None
+        if self.vehicle.is_capability_available(CapabilityId.PARKING_POSITION):
+            if self._positions().positions:  # pyright: ignore reportOptionalMemberAccess
+                return next(
+                    pos
+                    for pos in self._positions().positions  # pyright: ignore reportOptionalMemberAccess
+                    if pos.type == PositionType.VEHICLE
+                )
 
     @property
     def source_type(self) -> SourceType:  # noqa: D102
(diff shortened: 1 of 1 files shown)


custom_components/myskoda/device_tracker.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, coordinator: MySkodaDataUpdateCoordinator, vin: str) -> None:
         )
         super().__init__(coordinator, vin)
 
-    def _positions(self) -> Positions:
+    def _positions(self) -> Positions | None:
```
dvx76, line 47: Can't be `None` since we're raising an exception if it is?

custom_components/myskoda/device_tracker.py
```diff
@@ -68,17 +68,21 @@ def source_type(self) -> SourceType:  # noqa: D102
...
-        position = self._vehicle_position()
-        if position is None:
-            return None
-        return position.gps_coordinates.latitude
+        if self.vehicle.is_capability_available(CapabilityId.PARKING_POSITION):
+            position = self._vehicle_position()
+            if position is None:
+                return None
+            return position.gps_coordinates.latitude
+        return None
```
dvx76, line 76: nit: `None` is the default return value at the end of a function so this line isn't needed.

dvx76 (changes requested):

custom_components/myskoda/device_tracker.py
WebSpider: Yeah im kind of struggling with this one: When the in-car user disables position, the capability is present, not available and the _positions() is None

Prior99: I think we should return `None` if it's existing but unavailable and just raise an exception if the capability isn't existing at all. But we're checking the later one before creating the device tracker at all.

So in summary I think returning `None` would be better than raising an exception.

custom_components/myskoda/device_tracker.py
Prior99: I suggest to move the
```python
if self.vehicle.is_capability_available(CapabilityId.PARKING_POSITION):
```

Check into the `_vehicle_position` method instead and return `None`.

Prior99 (changes requested):

dvx76 (approved):

WebSpider:
A better implementation is in #93

PR #89, opened 2024-10-12
Fix km to m conversion for mqtt charged_range

### LINKED ISSUE #77: Electric car range goes to 0 (instead of unknown) during driving, version 1.4.5
**Describe the bug**
The range of the electric car should probably be unavailable while driving, but instead it reports 0. The screenshot shows the device tracker too, to show the 0 coincides with driving.

Did dot check (driving ;-) )

**Latest working release**
If the problem started after updating to a new release, please state latest working release.

**Debug logs**

**Installation:**
 - Home Assistant variant: HAOS version 2024.10.1
 - Python version (when running Core or Supervised) [e.g. 3.7, 3.8]
 - Python library versions ["python3 -m pip list"]

**Additional context**
IMG_0893

dvx76:
The API reports in meters while MQTT charging events report in km.

Makes me wonder how this will work for cars/profiles set in imperial units...

Fixes #77

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -20,7 +20,7 @@ from myskoda.models.operation_request import OperationName, OperationStatus
 from myskoda.models.user import User
 from myskoda.mqtt import EventCharging, EventType
 
-from .const import DOMAIN, FETCH_INTERVAL_IN_MINUTES, API_COOLDOWN_IN_SECONDS
+from .const import API_COOLDOWN_IN_SECONDS, DOMAIN, FETCH_INTERVAL_IN_MINUTES
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


custom_components/myskoda/coordinator.py
```diff
@@ -127,11 +127,11 @@ async def _on_charging_event(self, event: EventCharging):
...
             status = vehicle.charging.status
 
-            status.battery.remaining_cruising_range_in_meters = data.charged_range
+            status.battery.remaining_cruising_range_in_meters = (
+                data.charged_range * 1000
+            )
             status.battery.state_of_charge_in_percent = data.soc
             status.state = data.state
-            status.state = data.state
-        self.async_set_updated_data
```
WebSpider, line 134: Why not set the update?

dvx76: This doesn't do anything, it's just the name of the method, without actually calling it. It think it's leftover from some refactoring.

WebSpider (approved):

PR #103, opened 2024-10-17
Fix error when ChargingStatus.State is absent/None

### LINKED ISSUE #72: ValueError: Sensor sensor.skoda_enyaq_charging_state provides state value 'none', which is not in the list of options provided
In version 1.4.5 following error is logged.

````
Logger: homeassistant.components.sensor
Bron: helpers/entity_platform.py:724
integratie: Sensor (documentatie, problemen)
Eerst voorgekomen: 11:47:49 (1 gebeurtenissen)
Laatst gelogd: 11:47:49

myskoda: Error on device update!
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 724, in _async_add_entity
    await entity.async_device_update(warning=False)
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1300, in async_device_update
    await self.async_update()
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 571, in async_update
    await self.coordinator.async_request_refresh()
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 265, in async_request_refresh
    await self._debounced_refresh.async_call()
  File "/usr/src/homeassistant/homeassistant/helpers/debounce.py", line 114, in async_call
    await task
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 330, in async_refresh
    await self._async_refresh(log_failures=True)
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 453, in _async_refresh
    self.async_update_listeners()
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 168, in async_update_listeners
    update_callback()
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 533, in _handle_coordinator_update
    self.async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1005, in async_write_ha_state
    self._async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1130, in _async_write_ha_state
    self.__async_calculate_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1067, in __async_calculate_state
    state = self._stringify_state(available)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1011, in _stringify_state
    if (state := self.state) is None:
                 ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/sensor/__init__.py", line 636, in state
    raise ValueError(
ValueError: Sensor sensor.skoda_enyaq_charging_state provides state value 'none', which is not in the list of options provided

### PR DESCRIPTION
**Pull request by WebSpider**
Fixes #72

### DIFF
--- a/custom_components/myskoda/sensor.py
+++ b/custom_components/myskoda/sensor.py
@@ -295,6 +295,11 @@ class ChargingState(ChargingSensor):
                 return "mdi:power-plug-battery"
         return "mdi:power-plug"
 
+    @property
+    def available(self):  # noqa: D102
+        if status := self._status():
+            return status.state is not None
+
 
 class RemainingChargingTime(ChargingSensor):
     """Estimation on when the vehicle will be fully charged."""
--- a/custom_components/myskoda/sensor.py
+++ b/custom_components/myskoda/sensor.py
@@ -284,7 +284,8 @@ class ChargingState(ChargingSensor):
     @property
     def native_value(self) -> str | None:  # noqa: D102
         if status := self._status():
-            return str(status.state).lower()
+            if status.state:
+                return str(status.state).lower()
 
     @property
     def icon(self) -> str:  # noqa: D102
@@ -294,11 +295,6 @@ class ChargingState(ChargingSensor):
             if status.state == charging.ChargingState.CHARGING:
                 return "mdi:power-plug-battery"
         return "mdi:power-plug"
-
-    @property
-    def available(self):  # noqa: D102
-        if status := self._status():
-            return status.state is not None
 
 
 class RemainingChargingTime(ChargingSensor):

### CODE REVIEW

File: custom_components/myskoda/sensor.py [hunk 1]
```diff
@@ -295,6 +295,11 @@ def icon(self) -> str:  # noqa: D102
                 return "mdi:power-plug-battery"
         return "mdi:power-plug"
 
+    @property
+    def available(self):  # noqa: D102
+        if status := self._status():
+            return status.state is not None
```
dvx76, line 301: This will never evaluate to `False` because the `if` condition won't be entered if `self._status()` effectively returns `None`. In that case the function will end up returning `None` (the default) instead of a bool.

I think we just need `return self._status() is not None`

dvx76: Ah I read over the `status.state`

dvx76: I  guess just add a default `return False` at the end and we're good

```
    def available(self) -> bool:  # noqa: D102
        if status := self._status():
            return status.state is not None
        return False
```

Do we even need this if we fix `native_value()` so it doesn't return a stringified lowercase `None`? :)

```
    def native_value(self) -> str | None:  # noqa: D102
        if status := self._status():
            if status.state:
                return str(status.state).lower()
```

WebSpider: Yep, thats an equally good solution

sonar98: I will test with this line of code tonight

sonar98: This works in my setup:

````    
def native_value(self) -> str | None:  # noqa: D102
        if status := self._status():
            if status.state:
                return str(status.state).lower()

PR #110, opened 2024-10-20
Update entity names and IDs to include device name

### LINKED ISSUE #57: Add VIN number and/or integration name to ID of sensors
Currently all sensors have no link to the device and/or the integration. Example is that the milage sensor is sensor.milage and door sensor is binary_sensor.doors. It would be better if the sensors get a link to the skodaconnect integration or to the device (car) itself. So I would suggest one of these sensor names:

- sensor.skodaconnect.milage
- sensor.vin_number.milage

dvx76:
This will cause entity names and IDs to include the vehicle name, making it easier to identify and use the entities, especially when having multiple vehicles. The unique_id does not change.

E.g.

```
entity_description.key = charger_connected
unique_id = TMBXXXXXXXXXXXX_charger_connected
entitiy_id = sensor.skoda_enyaq_charger_connected
name = Skoda Enyaq Charger Connected
```

Reference: https://developers.home-assistant.io/docs/core/entity/#has_entity_name-true-mandatory-for-new-integrations

Fixes #57

This will be a somewhat breaking change in that in order to benefit from the new entitiy names users will need to delete and recreate the integration.

Thanks to @joostlek for the pointer!

### DIFF
--- a/custom_components/myskoda/entity.py
+++ b/custom_components/myskoda/entity.py
@@ -15,6 +15,7 @@ class MySkodaEntity(CoordinatorEntity):
 
     vin: str
     coordinator: MySkodaDataUpdateCoordinator
+    _attr_has_entity_name = True
 
     def __init__(
         self,


joostlek:
FYI, in core we don't see this as breaking change

WebSpider (approved):
Lgtm, this has been sufficiently discussed.

We should make sure to put this one in the release text, so that hacs users see this explicitly when they look at the release before hitting the update button

PR #114, opened 2024-10-21
Migrate Entity icons to icons.json file

WebSpider:
To better align with https://developers.home-assistant.io/blog/2024/01/19/icon-translations

Also made the URL render entity a diagnostic entity, no ordinary user would need to see this on a daily basis.


dvx76 (approved):
Nice!

Issue #42, opened 2024-09-27
Can't activate air Conditioning over UI


kotarokun:
Hello,
I want to change Air conditioning via UI but it always stand on auto. So I can only switch off if it is heating.
But I can switch via climate.turn_on and it works. 

Maybe just an switch to the actually state? So it stand off and can switch to auto?

Thanks for your help

WebSpider:
What version of HA and what version of the integration are you running?

WebSpider:
Also, are there any additional parameters you pass to `climate.turn_on`?

kotarokun:
I use actually 2024.06.4
And climate.turn_on works an I see in app it will send and after a while it will preheating the car.
I just add the Enyaq climate-entity. W
I be home in two days and maybe I have a look before

WebSpider:
Thank you. I will check on 2024.06 and see if i can reproduce this.

lightheaded:
For reference, I have two cars - Skoda Superb and Nissan Leaf. Nissan's climate works as expected, but I can't turn on my Škoda's climate entity via the UI, as it's seemingly always "on". I'm using latest HA and the addon.

In the actual "off" state:

Skoda:
![image](https://github.com/user-attachments/assets/69577533-ca97-4a88-840c-023101c8b23f)
![image](https://github.com/user-attachments/assets/518a6657-35b8-4135-8d64-684e20fc3dff)

Nissan:
![image](https://github.com/user-attachments/assets/77a1122e-d4ba-4cfe-b4d0-76421bbb006e)
![image](https://github.com/user-attachments/assets/d68d3dd4-fc5d-4d2f-b28a-851e190e97ec)

And in "on" state (i.e. climate manually started from the app):

Skoda:
![image](https://github.com/user-attachments/assets/ef9f21e5-e6f9-46bd-8ed9-b8347b46c3e8)
![image](https://github.com/user-attachments/assets/204b4290-b8f5-4939-97d5-54656fa046c6)

Nissan:
![image](https://github.com/user-attachments/assets/82a2f86e-1743-4dc8-9082-78633d708805)
![image](https://github.com/user-attachments/assets/b9574ec4-31c8-4468-8f08-d8ec8fbaddac)

lightheaded:
I am using this script as a workaround, as it toggles based climate on the `hvac_action` attribute:

```yaml
alias: Toggle Skoda Climate
sequence:
  - choose:
      - conditions:
          - condition: state
[...]

PR #117, opened 2024-10-21
Fix being unable to turn on HVAC from webinterface

WebSpider:
Related to #42

### DIFF
--- a/custom_components/myskoda/climate.py
+++ b/custom_components/myskoda/climate.py
@@ -69,13 +69,13 @@ class MySkodaClimate(MySkodaEntity, ClimateEntity):
 
     @property
     def hvac_modes(self) -> list[HVACMode]:  # noqa: D102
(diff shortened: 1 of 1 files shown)


dvx76 (approved):
Have you been able to test this?

WebSpider:
No not yet, ill throw it on my dev install tonight and merge if it works

WebSpider:
There was a bug left in the parsing of the incoming state that has now been fixed.

New behaviour is as follows:

1. When you change the temperature on the climate card, the target temperature of the AC is set, nothing else changes
2. When you turn off the AC in the climate card, the AC is forced off if it is running
3. When you turn the AC to HEAT/COOL in the climate card, the AC is forced ON and will start heating/cooling the car to the desired temperature

TODO: It would be really nice if we could mark the climate entity as read-only if the car is processing a change. I will open a new issue for that.

PR #123, opened 2024-10-22
Improve robustness of processing updates from MySkoda

WebSpider:
- Adds detection of regular polls failing and raises `UpdateFailed` to let HA handle this
- Adds detection of update requests following MQTT updates failing and lets HA handle this as well

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -80,7 +80,7 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
         myskoda.subscribe(self._on_mqtt_event)
 
     async def _async_update_data(self) -> State:
-        _LOGGER.debug("Perfoming scheduled update of all data for vin %s", self.vin)
+        _LOGGER.debug("Performing scheduled update of all data for vin %s", self.vin)
         try:
             vehicle = await self.myskoda.get_vehicle(self.vin)
             user = await self.myskoda.get_user()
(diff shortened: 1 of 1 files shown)


custom_components/myskoda/coordinator.py
```diff
@@ -79,8 +80,12 @@ def __init__(
         myskoda.subscribe(self._on_mqtt_event)
 
     async def _async_update_data(self) -> State:
-        vehicle = await self.myskoda.get_vehicle(self.vin)
-        user = await self.myskoda.get_user()
+        _LOGGER.debug("Performing scheduled update of all data for vin %s", self.vin)
+        try:
+            vehicle = await self.myskoda.get_vehicle(self.vin)
+            user = await self.myskoda.get_user()
+        except ClientError as err:
```
dvx76, line 87: out of scope here but I wonder if our myskoda API package should raise its own exceptions instead of letting `aiohttp` (and other) exceptions bubble up.

custom_components/myskoda/coordinator.py
```diff
@@ -79,8 +80,12 @@ def __init__(
...
+            raise UpdateFailed("Error getting update from MySkoda API: %s", err)
```
dvx76, line 88: I think it's pylint (which I use at work) which always yells about me about raising `from` the original exception.

```suggestion
            raise UpdateFailed("Error getting update from MySkoda API: %s", err) from err
```

custom_components/myskoda/coordinator.py
WebSpider: Would be nice, but i checked HA's coordinator code and in the end they catch aiohttp's exceptions as well as stopgap

custom_components/myskoda/coordinator.py
WebSpider: Was just following the examples about calling UpdateFailed and it passed the linter 😂

dvx76: Yeah it's not a must for me. It's this one: https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/raise-missing-from.html

dvx76 (approved):

PR #153, opened 2024-10-31
Add current battery capacity

WebSpider:
This adds a sensor and statistics so you can see the development of your battery capacity over time.

This seems interesting, as I noticed that this has diminished over time for my car from 82 kWh to 77 kWh already


WebSpider:
I will connect my OBD later today to make sure :+1:

WebSpider:
Yep, you are right, my actual battery state is not equal to 77 (in fact, a bit lower..)

Issue #166, opened 2024-11-03
Superb iV: Internal Server Error for BatteryCare and ChargeLimit functionality 


prvakt:
Receiving Internal Server Error when testing set-charge-limit and set-care-mode.. Probably because its not supported by my car? I can't find CareMode anywhere in MySkoda app, car infotaiment or in old integration. For charge limit, also it seems its not available in MySkoda or car infotaiment (old integration has support for this feature but I have never tested it)

**set-charge-limit**
```
Logger: homeassistant.helpers.script.websocket_api_script
Source: helpers/script.py:525
First occurred: 12:34:50 PM (1 occurrences)
Last logged: 12:34:50 PM

websocket_api script: Error executing script. Unexpected error for call_service at pos 1: 500, message='Internal Server Error', url='https://mysmob.api.connect.skoda-auto.cz/api/v1/charging/XXXXXXXXXXXXXXXXX/set-charge-limit'
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 525, in _async_step
    await getattr(self, handler)()
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 763, in _async_call_service_step
    response_data = await self._async_run_long_action(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 726, in _async_run_long_action
    return await long_task
           ^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2761, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2804, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 996, in entity_service_call
    single_response = await _handle_entity_call(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1068, in _handle_entity_call
    result = await task
             ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/number/__init__.py", line 122, in async_set_value
    await entity.async_set_native_value(native_value)
  File "/config/custom_components/myskoda/number.py", line 75, in async_set_native_value
[...]

PR #167, opened 2024-11-03
Add forbidden capabilities

WebSpider:
Forbidden capabilities are capabilities that cannot be present for a sensor to be added.
They are effectively the reverse of required capabilities, that must be present for a sensor to be added.

The logic added in this PR is follows:

- A sensor will be added when the vehicle has *none* of the capabilities in forbidden capabilities and *all* of the capabilities in required capabilities

Example here is the battery-care switch for a Superb iV:
The Superb iV does not have the function, but it does support the CHARGING capability.

When you add the CHARGING_MQB capability to the forbidden capabilities, the battery-care switch will now never be added for the Superb iV:

### DIFF
--- a/custom_components/myskoda/entity.py
+++ b/custom_components/myskoda/entity.py
@@ -46,9 +46,17 @@ class MySkodaEntity(CoordinatorEntity):
     def required_capabilities(self) -> list[CapabilityId]:
         return []
 
+    def forbidden_capabilities(self) -> list[CapabilityId]:
+        return []
+
     def is_supported(self) -> bool:
         return all(
             self.vehicle.has_capability(cap) for cap in self.required_capabilities()
+        )
+
+    def is_forbidden(self) -> bool:
+        return all(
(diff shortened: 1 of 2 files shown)


WebSpider:
This is related to #166

dvx76 (changes requested):
While it looks like this will be usable for the specific case of #166 , I do wonder if this isn't oversimplifying things. Wouldn't it be conceivable that _some_ models with the `CHARGING_MQB` capability support the charging limit, while other don't?

Wouldn't it be better to use the `errors` in the response from the charging (and I guess others) endpoint, which explicitly conveys the information we need?

E.g. from the suber_iv fixture:

```
    errors:
    - description: Care mode is not available.
      type: CARE_MODE_IS_NOT_AVAILABLE
    - description: Auto unlock is not available.
      type: AUTO_UNLOCK_IS_NOT_AVAILABLE
    - description: Charge limit is not available.
      type: CHARGE_LIMIT_IS_NOT_AVAILABLE
```

dvx76 (approved):

PR #189, opened 2024-11-06
Configurable poll interval

WebSpider:
This sets up a configurable poll interval as an optional configuration. If unset, it defaults to 30 minutes.


dvx76 (approved):
Thanks for adding this! I will definitely be using this myself until I get proper soc events for every % charged.

Issue #175, opened 2024-11-04
API error


SenadM:
all my entities are unavailable. Log below. 

This error originated from a custom integration.

Logger: custom_components.myskoda
Source: custom_components/myskoda/__init__.py:41
integration: MySkoda (documentation, issues)
First occurred: 18:52:05 (1 occurrences)
Last logged: 18:52:05

Login with MySkoda failed.
Traceback (most recent call last):
  File "/config/custom_components/myskoda/__init__.py", line 41, in async_setup_entry
    await myskoda.connect(config.data["email"], config.data["password"])
  File "/usr/local/lib/python3.12/site-packages/myskoda/myskoda.py", line 123, in connect
    user = await self.get_user()
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/myskoda/myskoda.py", line 262, in get_user
    return (await self.rest_api.get_user(anonymize=anonymize)).result
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/myskoda/rest_api.py", line 225, in get_user
    data=await self._make_get_request(url),
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/myskoda/rest_api.py", line 95, in _make_get_request
    return await self._make_request(url=url, method="GET")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/myskoda/rest_api.py", line 85, in _make_request
    response.raise_for_status()
  File "/usr/local/lib/python3.12/site-packages/aiohttp/client_reqrep.py", line 1121, in raise_for_status
    raise ClientResponseError(
aiohttp.client_exceptions.ClientResponseError: 500, message='Internal Server Error', url='https://mysmob.api.connect.skoda-auto.cz/api/v1/users'

dvx76:
Yeah there was some downtime on the API today apparently. @SenadM please confirm if you're still having issues.

> Would be great if the sensors wouldn’t go to unavailable if the Skoda servers are down, but just keep the last values known.

[...]

PR #221, opened 2024-11-08
Implementation of door lock for supporting models

WebSpider:
This will be added to a prerelease of the integration, since I cannot test it myself (no cars that support locking)

Mode of operation of this lock:

- If you have the correct capabilties in your car, the lock entity is created, but by default unavailable
- If you enter the S-PIN in optional settings (next to polling interval), the lock becomes available for operation


custom_components/myskoda/config_flow.py
```diff
@@ -43,9 +44,13 @@ async def validate_options_input(
...
-        if CONF_POLL_INTERVAL_MIN <= polling_interval <= CONF_POLL_INTERVAL_MAX:
-            return user_input
-        raise SchemaFlowError("invalid_polling_interval")
+        if not CONF_POLL_INTERVAL_MIN <= polling_interval <= CONF_POLL_INTERVAL_MAX:
+            raise SchemaFlowError("invalid_polling_interval")
+
+    if CONF_SPIN in user_input:
+        s_pin: str = user_input[CONF_SPIN]
+        if not s_pin.isdigit():
+            raise SchemaFlowError("invalid_spin_format")
```
dvx76, line 53: Do we want to validate the `s-pin` here with the `MySkoda.verify_spin()` from https://github.com/skodaconnect/myskoda/pull/175/files ?

dvx76 (commented):
LGTM but I also can't test and I guess you want to include the feedback from Discord (missing state I think)

custom_components/myskoda/config_flow.py
WebSpider: I think prefer to use `MySkoda.verify_spin()` when storing the S-PIN in config, instead of every time when calling a feature that requires the S-PIN

prvakt:
the inital state after reload of integration was correct but when I click lock (because vehicle was initialy unlocked), following error showed :
```
2024-11-09 09:28:59.662 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [140145328769920] Unexpected exception
File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 245, in handle_call_service
response = await hass.services.async_call(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/core.py", line 2761, in async_call
response_data = await coro
^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/core.py", line 2804, in _execute_service
return await target(service_call)
^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 996, in entity_service_call
single_response = await _handle_entity_call(
^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1068, in _handle_entity_call
result = await task
^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/components/lock/__init__.py", line 226, in async_handle_lock_service
await self.async_lock(**self.add_default_code(kwargs))
File "/config/custom_components/myskoda/lock.py", line 81, in async_lock
await self.coordinator.update_status()
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'MySkodaDataUpdateCoordinator' object has no attribute 'update_status'
```
but door was locked . just state remain as unlocked state until it was refreshed after some time together with all other sensors

prvakt:
seems like `self.coordinator.update_status` is not triggered from **lock.py**

in order to fix this please remove both `self.coordinator.update_status` lines from **lock.py** add following lines to **coordinator.py** `_on_operation_event` 
```
        if event.operation.operation in [
            OperationName.LOCK,
            OperationName.UNLOCK,
        ]:
            await self.update_vehicle()
```

WebSpider:
Thanks for testing this and providing the solution!

Please test with v1.8.2b2

prvakt:
> Thanks for testing this and providing the solution!
> 
> Please test with v1.8.2b2

13:57:27.520 Unlock triggered
13:57:38.782 Vehicle unlocked
13:58:19.370 Status updated

seems much better now, sensor update is working as expected, but I suggest to think about using `status":"COMPLETED_SUCCESS"` from MQTT in order to detect change of sensor state faster without polling car for update (at least in this case we can save 30s and 1 API request)

WebSpider:
This did not make it into 1.8.2, it can now be accessed as prerelease `v1.8.3b1`

WebSpider:
The lag between updating switch and the actual status is mostly caused by sluggish response from the car.
For now, having this feature outweighs it not being speedy enough for me, so this PR is ready for review :partying_face:

PR #229, opened 2024-11-10
Update translations for charging state connect_cable

### LINKED ISSUE #137: Enyaq: Charging State=Cable Disconnected while Charger Connected=Plugged in
I use HA to charge when the price are lowesst. So i plug the car in and the easee home charger is paused by setting dynamic max = 0 until it is time to charge.

That is the cable is plugged in, the charger is ready with the status PAUSED/AWAITING START

MySkoda shows correctly that Charger Connected=Plugged in but the Charging State=Cable Disconnected 
I do not now whether it just an bad translation or something less obvious

The old skoda connect interface shows 

Charging cable connected = Plugged in
Charging cable locked = Locked
External power = Off <<< Maybe that is similar to the Charging state in MySkoda ?

When the cars start to charge it changes to 
Charging State = Charging (Myskoda)
External Power = On (Skoda Connect)

Here from the small test:
image

Set dynamic on charger set > 6A to start charging and back to zero 8 minutes later

dvx76:
Fixes #137

We may have misunderstood what Skoda and their API mean with a charging status of `CONNECT_CABLE`

We've been assuming it means "You need to connect the cable" but instead it probably means "There is a cable connected but that's it, we're neither charging nor ready to charge".

A better translation is "No Power".

Note that the binary_sensor `charger_connected` already tells us exactly if a charge cable is connected or not.

Translations based on google translate ...

### DIFF
--- a/custom_components/myskoda/translations/cs.json
+++ b/custom_components/myskoda/translations/cs.json
@@ -140,7 +140,7 @@
                 "name": "Stav nabíjení",
                 "state": {
                     "charging": "Nabíjení",
-                    "connect_cable": "Kabel odpojen",
+                    "connect_cable": "Žádná síla",
                     "conserving": "Uchovávání nabití",
                     "ready_for_charging": "Připraveno k nabíjení"
                 }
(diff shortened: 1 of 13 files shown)


WebSpider (approved):
Makes sense
Let's see if we get any user feedback when we change this translation

Issue #212, opened 2024-11-08
Too many requests occurred during getting warning lights report for vehicle


equareo:
After running the service for some hours hours it ends up with a 429 and 24 hour ban. It always end with this exact get request getting 429. I've not done any toggles, just started the integration, and let it run. 

2024-11-08 07:41:15.220 DEBUG (MainThread) [myskoda.myskoda] Trace: GET https://mysmob.api.connect.skoda-auto.cz/api/v1/vehicle-heal - response: 429 (124 bytes) Too many requests occurred during getting warning lights report for vehicle {vinnumberremoved} and accept-language {en-US}.
2024-11-08 07:41:15.221 ERROR (MainThread) [myskoda.rest_api] Invalid status for GET request to /v1/vehicle-health-report/warning-lights/vinnumberremoved: 429
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/myskoda/rest_api.py", line 85, in _make_request
    response.raise_for_status()
  File "/usr/local/lib/python3.12/site-packages/aiohttp/client_reqrep.py", line 1121, in raise_for_status
    raise ClientResponseError(
aiohttp.client_exceptions.ClientResponseError: 429, message='Too Many Requests', url='https://mysmob.api.connect.skoda-auto.cz/api/v1/vehicle-health-report/warning-lights/vinnumberremoved'
2024-11-08 07:41:15.224 ERROR (MainThread) [custom_components.myskoda.coordinator] Error fetching myskoda data: ('Error getting update from MySkoda API: %s', ClientResponseError(RequestInfo(url=URL('https://mysmob.api.connect.skoda-auto.cz/api/v1/vehicle-health-report/warning-lights/vinnumberremoved'), 

Car: Citigo

Here is a link to a long log:  [https://bin.68degrees.no/?acf71efcad3c395a#E33sNjb8p2MXjRNsCcpPh2Xy7uU1S5vGE7mcEo7Z5kbZ](https://bin.68degrees.no/?acf71efcad3c395a#E33sNjb8p2MXjRNsCcpPh2Xy7uU1S5vGE7mcEo7Z5kbZ)

WebSpider:
Thanks for your report. The logfile contains useful information. 
We need to figure out what's happening and how to prevent this from happening. Looks like not a quick fix, unfortunately.

dvx76:
[...]

PR #261, opened 2024-11-12
Trigger proper exceptions inside HA when an update fails

WebSpider:
Currently HA throws an exception inside the logs when an update fails.

This makes HA handle the failure nicer


WebSpider:
Related to #212

PR #262, opened 2024-11-12
Rename lights to parkinglights

### LINKED ISSUE #245: Add Parking Light Status
This is a template issue from the project to get more SkodaConnect sensors into `homeassistant-myskoda`.

**Sensor**
Parking Light Status

**Last known status**
No support in MySkoda API's discovered yet

WebSpider:
Lights were incorrectly named, as in reality they represent parkinglights.

This removes the lights entity and introduces a new parkinglights entity without any functional change
Fixes #245

### DIFF
--- a/custom_components/myskoda/binary_sensor.py
+++ b/custom_components/myskoda/binary_sensor.py
@@ -226,7 +226,7 @@ class ParkingLightsOn(StatusBinarySensor):
     """Detects whether the parking-lights are on."""
 
     entity_description = BinarySensorEntityDescription(
-        key="parkinglights_on",
+        key="lights_on",
         device_class=BinarySensorDeviceClass.LIGHT,
         translation_key="parkinglights_on",
     )
--- a/custom_components/myskoda/binary_sensor.py
+++ b/custom_components/myskoda/binary_sensor.py
@@ -41,7 +41,7 @@ async def async_setup_entry(
             WindowsOpen,
             TrunkOpen,
             BonnetOpen,
-            LightsOn,
+            ParkingLightsOn,
             ChargerConnected,
             ChargerLocked,
             SunroofOpen,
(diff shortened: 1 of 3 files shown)


custom_components/myskoda/translations/en.json
```diff
@@ -54,20 +54,20 @@
...
-                }
-            },
             "locked": {
                 "name": "Vehicle Locked",
                 "state": {
                     "off": "Locked",
                     "on": "Unlocked"
                 }
             },
+            "parkinglights_on": {
```
dvx76, line 64: Other translations should be updated as well?

custom_components/myskoda/binary_sensor.py
```diff
@@ -222,13 +222,13 @@ def is_supported(self) -> bool:
         return False
 
 
-class LightsOn(StatusBinarySensor):
-    """Detects whether the lights are on."""
+class ParkingLightsOn(StatusBinarySensor):
```
dvx76, line 225: I guess we'll need to call this out in the release notes so people know the `lights_on` sensor is gone.

I imagine HA has build in mechanisms to 'migrate' things but at this stage in the development I'm ok to just ask users to handle it 😅

custom_components/myskoda/translations/en.json
WebSpider: Yes, once it is pushed to main, it will be added as a translatable string in Crowdin, where people can translate it without having to know git :grin:

custom_components/myskoda/binary_sensor.py
WebSpider: SemVer tells us this is 2.0.0 then :open_mouth:

WebSpider: I could make it a bit less intrusive and only update icon and translation. That may be a better way forward.

dvx76 (approved):

PR #265, opened 2024-11-12
Honk + Flash + Range sensors

prvakt:
Added Honk & Flash button
Added separate Combustion and Electric range sensors (for hybrid)
Added Fuel level (for hybrid and ICE)
Changed range to use value from driving_range.total_range_in_km instead of a value from charging (so it can be used by all engine types)


WebSpider:
Sorry, you need to split this up into multiple PR's

One PR per subject

PR #268, opened 2024-11-12
Change range to work with all car types, make icon car-type dependant

prvakt:
Changed range to use value from driving_range.total_range_in_km instead of a value from charging (so it can be used by all engine types)

### DIFF
--- a/custom_components/myskoda/sensor.py
+++ b/custom_components/myskoda/sensor.py
@@ -176,9 +176,11 @@ class RemainingDistance(ChargingSensor):
 
     @property
     def native_value(self) -> int | float | None:  # noqa: D102
-        if status := self._status():
-            if status.battery.remaining_cruising_range_in_meters is not None:
-                return status.battery.remaining_cruising_range_in_meters / 1000
+        if range := self.vehicle.driving_range:
+            return range.total_range_in_km
+            
+    def required_capabilities(self) -> list[CapabilityId]:
+        return [CapabilityId.STATE]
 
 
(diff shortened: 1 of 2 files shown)


custom_components/myskoda/sensor.py
```diff
@@ -176,9 +176,11 @@ class RemainingDistance(ChargingSensor):
 
     @property
     def native_value(self) -> int | float | None:  # noqa: D102
-        if status := self._status():
-            if status.battery.remaining_cruising_range_in_meters is not None:
-                return status.battery.remaining_cruising_range_in_meters / 1000
+        if range := self.vehicle.driving_range:
```
dvx76, line 179: nit: maybe we should rename this class to `Range` now? And update the doc-string accordingly.

dvx76 (commented):
Great. No idea why we didn't use the range from the driving_range right away 🤷‍♂️ 

Minor comment + linter wants an extra newline (tip: install https://docs.astral.sh/ruff/ in your editor and it will auto-format things for you right away)

custom_components/myskoda/sensor.py
WebSpider: Also its not a charging sensor any more, chargingsensors only show up on cars with charging capability

WebSpider (approved):
Looks good to me. As soon as you fix the merge conflict, we can go ahead and commit this. Thanks for your contribution!

WebSpider (approved):

WebSpider (approved):
Thanks for your contribution!

Issue #275, opened 2024-11-14
Remove charging sensors which are not supported


prvakt:
Following sensors are not available in Superb iV and showing `UNKNOWN` state
`sensor.*_charging_power`
`sensor.*_charging_rate`
`sensor.*_target_battery_percentage`
```
{
    "isVehicleInSavedLocation": false,
    "status": {
        "remainingTimeToFullyChargedInMinutes": 190,
        "state": "CHARGING",
        "chargeType": "AC",
        "battery": {
            "remainingCruisingRangeInMeters": 22000,
            "stateOfChargeInPercent": 67
        }
    },
    "settings": {
        "availableChargeModes": [],
        "maxChargeCurrentAc": "REDUCED"
    },
    "carCapturedTimestamp": "2024-11-14T09:53:19Z",
    "errors": [
        {
            "type": "CARE_MODE_IS_NOT_AVAILABLE",
            "description": "Care mode is not available."
        },
        {
            "type": "AUTO_UNLOCK_IS_NOT_AVAILABLE",
            "description": "Auto unlock is not available."
        },
        {
            "type": "CHARGE_LIMIT_IS_NOT_AVAILABLE",
            "description": "Charge limit is not available."
        }
    ]
}
```
just don't know which capability can be chosen to restrict those sensors - maybe CHARGING_MEB or EXTENDED_CHARGING_SETTINGS? those are not available in my Superb iV

WebSpider:
It's probably CHARGING_MEB.

prvakt:
ok I will create a PR for this

WebSpider:
CARE_MODE relates to a switch, which was made limited to `BATTERY_CHARGING_CARE` in c5e1dd31.

Please make sure you remove orphaned entities first ;)

PR #276, opened 2024-11-14
restrict some charge sensors for cars with CHARGING_MQB capability

prvakt:


prvakt:
to fix #275

custom_components/myskoda/sensor.py
```diff
@@ -169,6 +169,9 @@ def native_value(self) -> float | None:  # noqa: D102
         if status := self._status():
             return status.charge_power_in_kw
 
+    def required_capabilities(self) -> list[CapabilityId]:
+        return [CapabilityId.CHARGING_MEB]
+
```
WebSpider, line 174: If you only know they are not present in one type of vehicle, it would be better to use `forbidden_capabilities` to make sure the sensor does not show up for your type of car. I think it's `CHARGING_MQB`

WebSpider (changes requested):

WebSpider (approved):
Let's go!

Issue #209, opened 2024-11-08
Add repository to HACS


tigers75:
I know the add-in is still in beta, but with the announced retire of the SkodaConnect add-in I think this should be added to HACS for an easier replacement.

tigers75:
My bad, I didn't notice it's already there.

PR #278, opened 2024-11-14
Handle new errors observed

WebSpider:
Add handling for 

- Failed authentications
- New terms and conditions
- Errors seen when service had problems

Depends on yet-to-be-released `myskoda` library with skodaconnect/myskoda#209 merged


custom_components/myskoda/__init__.py
```diff
@@ -41,8 +49,17 @@ async def async_setup_entry(hass: HomeAssistant, config: ConfigEntry) -> bool:
 
     try:
         await myskoda.connect(config.data["email"], config.data["password"])
+    except AuthorizationFailedError:
+        _LOGGER.debug("Authorization with MySkoda failed.")
+        raise ConfigEntryAuthFailed("Authentication failed.")
+    except TermsAndConditionsError:
+        _LOGGER.debug("New terms and conditions detected in login flow.")
```
dvx76, line 56: Is that all the user gets, or does the re-raise make it pop up somewhere else? Do we want to be a bit more verbose here and provide instructions (_log out and back in again on the MySkoda app to access the new T&C_)?

dvx76: Also what determines whether we raise a new HA-specific exception, or re-raise the same exception, or return False?

dvx76 (approved):

custom_components/myskoda/__init__.py
WebSpider: > Is that all the user gets, or does the re-raise make it pop up somewhere else? Do we want to be a bit more verbose here and provide instructions (_log out and back in again on the MySkoda app to access the new T&C_)?

I have not yet figured out how to make this a proper message in the integration status that also is translatable etc.

I will make the message a bit more elaborate, for instance adding an error level logger as well, while I figure out how I can properly inform the user in the UI as well.

WebSpider: > Also what determines whether we raise a new HA-specific exception, or re-raise the same exception, or return False?

You make it sounds like there is a determination there :grin:

I will make it a bit more consistent.

PR #305, opened 2024-11-17
Let Range fall back to battery if driving_range is missing

### LINKED ISSUE #302: Range entity no longer working
**Describe the bug**
Since the update to 1.9.0 I noticed I no longer have a valid range value.
I'm using the integration with my Skoda Enyaq IV

Yes I can see the range in the MySkoda App

**Latest working release**
Pre 1.9.0

**Debug logs**

**Installation:**
Core: 2024.11.2
Supervisor: 2024.11.2
Operating System: 13.2
Frontend: 20241106.2

image

dvx76:
Due to missing subscriptions for the STATE capability the driving_range can be missing while the battery data is available.

Fixes #302

### DIFF
--- a/custom_components/myskoda/sensor.py
+++ b/custom_components/myskoda/sensor.py
@@ -18,12 +18,14 @@ from homeassistant.const import (
 )
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
-from homeassistant.helpers.typing import DiscoveryInfoType  # pyright: ignore [reportAttributeAccessIssue]
+from homeassistant.helpers.typing import (
+    DiscoveryInfoType,  # pyright: ignore [reportAttributeAccessIssue]
+)
 
 from myskoda.models import charging
 from myskoda.models.charging import Charging, ChargingStatus
-from myskoda.models.info import CapabilityId
 from myskoda.models.driving_range import EngineType
+from myskoda.models.info import CapabilityId
 
 from .const import COORDINATORS, DOMAIN
 from .entity import MySkodaEntity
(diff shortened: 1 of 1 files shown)


custom_components/myskoda/sensor.py
```diff
@@ -263,8 +263,10 @@ def native_value(self) -> int | float | None:  # noqa: D102
         if range := self.vehicle.driving_range:
             return range.total_range_in_km
 
-    def required_capabilities(self) -> list[CapabilityId]:
```
WebSpider, line 266: Removing the `required_capabilities` will cause the entity to always be created. Wouldn't it be better here to return the relevant capability we have, such as 

```python
if self.vehicle.has_capability(CapabilityId.CHARGING):
  return [CapabilityId.CHARGING]
elif self.vehicle.has_capability(CapabilityId.XXX)
   etc
else
  return [CapabilityId.STATE]
   ```

dvx76: Can do. Starts to feel like it might make more sense to fully delegate the decision of 'being supported' to each entity class itself. E.g every `MySkodaEntity` would have a `is_supported` method (base class just returns `True`). Most would just inspect if one of more capabilities are present (or missing, in case of the forbidden_capabilities) but more advanced logic would then be possible. I also like how things are set up in the BMW integration: https://github.com/home-assistant/core/blob/c154ac26eb6287fa144cd48cbe8fd80bbe4fb09d/homeassistant/components/bmw_connected_drive/sensor.py#L58

Might take a look at that later. But out of scope here.

WebSpider: I like that idea of the `is_supported` class, it should make this cleaner.

About the BMW integration: It has stuff I like, and stuff I like less. Let's pick the right things to re-use :-)

WebSpider: As also just discussed in Discord, it would probably be better to override `is_supported()` in this class to something else than "just" the capabilities, but also take vehicle state into account.

dvx76: Updated

PR #319, opened 2024-11-18
Introduce readonly mode

WebSpider:
In read-only mode, you can only read from your car, but never write to it.
Read-only mode defaults to "off" and you can turn it on in the options.

The following entities are unavailable:

- All buttons (honk/flash)
- All numbers (charge level)
- All switches (window heating, charge on/off, etc)
- All climate entities are available, but have no options to get turned off/on


dvx76:
Would this effectively delete those entities when enabling read-only on an existing install?

dvx76 (approved):

WebSpider:
Will do a real-world check so we can document properly.

PR #337, opened 2024-11-21
Remove DURATION type sensor due to rendering issues

WebSpider:
When setting the sensor as a DURATION (which it factually is), homeassistant frontend will always render the value in hours, not days.

According to https://github.com/home-assistant/frontend/issues/18151 it is currently not possible to fix this in another way, so pragmatically, this is the way to inconvenience users the least.

### DIFF
--- a/custom_components/myskoda/sensor.py
+++ b/custom_components/myskoda/sensor.py
@@ -312,7 +312,8 @@ class InspectionInterval(MySkodaSensor):
 
     entity_description = SensorEntityDescription(
         key="inspection",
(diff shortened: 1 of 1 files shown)


dvx76 (approved):
Isn't this trading one inconvenience for another / impacting one set of users to satisfy another? With no device_class cards which correctly do convert the value will no longer be able to? Personally I don't use these sensors so if this is what people want I'm fine with it 😄

WebSpider:
Not convinced this is the best solution either. I'll think on it.

PR #345, opened 2024-11-24
Add button to generate fixtures into HA log

prvakt:


prvakt:
this feature will simplify retrieval of fixtures for users as they can easily trigger fixture generation from HA and response is traced as a debug message in HomeAssistant log

dvx76:
Can we include instructions in the README to cover what this is for and how to retrieve the results.

What does this look like in practice? Is it achievable to retrieve the fixture from the logs? I know some of these fixture files are hundreds of lines long. Is it in the right format?

prvakt:
> Can we include instructions in the README to cover what this is for and how to retrieve the results.
> 
> What does this look like in practice? Is it achievable to retrieve the fixture from the logs? I know some of these fixture files are hundreds of lines long. Is it in the right format?

Yeah the trace is available in the log as a single trace.. In my case it was around 32kB of data which was very easy to copy out from log viewer.. I will update README tomorrow accordingly.

dvx76 (approved):
LGTM. Let's see how it goes in the hands of the users 😄

rgjertsen:
For translation purposes. I am not sure what it means to generate fixtures, so I'm not sure how to translate it😅

PR #355, opened 2024-11-27
add AC settings switches

prvakt:
- AC when Unlocked (Enyaq)
- AC without External Power (Superb iV, Citigo iV)
- Left Seat Heating with AC (Enyaq)
- Right Seat Heating with AC (Enyaq)
- Window Heating with AC (Enyaq)

+ moved ReducedCurrent and BatteryCareMode switch into EntityCategory.CONFIG together with 5 new

 AC without External Power tested and working, other switches must be tested as my car doesn't support those features

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,12 @@ If your desired language is not available, please [open an issue](https://github
 - Reduced Current
 - Battery Care Mode
 - Doors
+- AC when Unlocked
+- AC without External Power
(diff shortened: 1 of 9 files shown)


WebSpider (approved):
LGTM.

In general it is nicer to split types of changes (new sensors vs changes of sensors) into seperate PR's, but this is OK for now.

PR #357, opened 2024-11-27
AC changed to use only AIR_CONDITIONING capability 

prvakt:
should be sufficient because Citigo iV doesnt have the other capability (AIR_CONDITIONING_SAVE_AND_ACTIVATE) but still can enable/disable AC from myskoda app


WebSpider (approved):
lgtm

Issue #356, opened 2024-11-27
AttributeError: 'ServiceEventData' object has no attribute 'charged_range'


prvakt:
with latest version of myskoda (`1.11.1`) I got following error

```
Logger: homeassistant
Source: custom_components/myskoda/coordinator.py:171
integration: MySkoda ([documentation](https://github.com/skodaconnect/homeassistant-myskoda/blob/main/README.md), [issues](https://github.com/skodaconnect/homeassistant-myskoda/issues))
First occurred: 12:02:35 PM (1 occurrences)
Last logged: 12:02:35 PM

Error doing job: Task exception was never retrieved (None)
Traceback (most recent call last):
  File "/config/custom_components/myskoda/coordinator.py", line 117, in _on_mqtt_event
    await self._on_charging_event(event)
  File "/config/custom_components/myskoda/coordinator.py", line 171, in _on_charging_event
    if data.charged_range is not None:
       ^^^^^^^^^^^^^^^^^^
AttributeError: 'ServiceEventData' object has no attribute 'charged_range'
```
after receiving service-event
```
2024-11-27 12:02:35.545 DEBUG (MainThread) [myskoda.mqtt] Message (service-event) received for XXXXXXXXXXXXXXXXX on topic charging: b'{"version":1,"traceId":"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx","timestamp":"2024-11-27T11:02:35Z","producer":"SKODA_MHUB","name":"charging-status-changed","data":{"userId":"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx","vin":"XXXXXXXXXXXXXXXXX"}}'
```

fursov:
@WebSpider just to clarify: when I mentioned the https://github.com/skodaconnect/myskoda/pull/235, I meant the HA component now needs adaptation, this is not a bug in the library. This now happens because the interface towards library has changed.

fursov:
I must admit that I have not taken care of that myself and also forgot to inform you about potential issue.

WebSpider:
> I must admit that I have not taken care of that myself and also forgot to inform you about potential issue.

Don't worry about it, it's why this piece of software has a big disclaimer :-)

Should you change the interface of the library, next time please just add this in the PR, so we can check accordingly.
As reviewers we also learned to check on interface changes better.

reineo:
[...]

PR #359, opened 2024-11-27
Fix missing field error for charging event

fursov:
This is draft change for #356.

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -21,6 +21,7 @@ from myskoda.event import (
     ServiceEventTopic,
 )
 from myskoda.models.operation_request import OperationName, OperationStatus
+from myskoda.models.service_event import ServiceEventChangeSoc
 from myskoda.models.user import User
 from myskoda.mqtt import EventCharging, EventType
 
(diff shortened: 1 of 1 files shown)


fursov:
Seems this works for the event `charging-status-changed`. My Superb does not fire `change-soc` event, would be nice of this is also tested against that event.

fursov:
@WebSpider there are two options how this fix can be done:
- one is to use isinstance to check which class fires the event (done in the current change)
- two is to use event name to distinguish the data structure
What would you prefer? I slightly incline towards event name, but then the event class is not visible here anymore, we would just implicitly assume the fields in the data...

custom_components/myskoda/coordinator.py
```diff
@@ -168,27 +169,27 @@ async def _on_charging_event(self, event: EventCharging):
...
                 )
-            if data.soc is not None:
                 status.battery.state_of_charge_in_percent = data.soc
-            if data.time_to_finish is not None:
-                status.remaining_time_to_fully_charged_in_minutes = data.time_to_finish
-            if data.state is not None:
+                if data.time_to_finish is not None:
+                    status.remaining_time_to_fully_charged_in_minutes = (
+                        data.time_to_finish
+                    )
```
WebSpider, line 180: Please format this using `match` and `case`, check https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching

We are python3.10+ (since HA is 3.12 and going to 3.13)

You could try to match the instance type

WebSpider (changes requested):
Thanks for your contribution!

Let's try and keep the classes in place unless it really complicates things. Using classes makes it really clear what type of messages we are passing around.

custom_components/myskoda/coordinator.py
fursov: Done

custom_components/myskoda/coordinator.py
```diff
@@ -168,38 +169,35 @@ async def _on_operation_event(self, event: EventOperation) -> None:
...
 
     async def _on_charging_event(self, event: EventCharging):
-        # TODO @webspider: Refactor with proper classes
         vehicle = self.data.vehicle
-
         data = event.event.data
 
         if vehicle.charging is None or vehicle.charging.status is None:
             await self.update_charging()
         else:
```
WebSpider, line 177: offtopic: I actually don't think this should be an `else` structure. If we don't have charging data, we should populate it and continue. Maybe that fits better in another PR.

custom_components/myskoda/coordinator.py
```diff
@@ -168,38 +169,35 @@ async def _on_operation_event(self, event: EventOperation) -> None:
...
+                    status = vehicle.charging.status
+                    status.battery.remaining_cruising_range_in_meters = (
+                        data.charged_range * 1000
+                    )
+                    status.battery.state_of_charge_in_percent = data.soc
+                    if data.time_to_finish is not None:
+                        status.remaining_time_to_fully_charged_in_minutes = (
+                            data.time_to_finish
+                        )
+                        status.state = data.state
```
WebSpider, line 190: Do we want to set `last_updated` to `timestamp` here?

WebSpider (commented):
Overall a nice bugfix that lays some good groundwork!
One question left, put inline.

custom_components/myskoda/coordinator.py
fursov: Yes, to be handled in #370

prvakt:
tested with my Superb and looks OK.. also #368 work correctly when I used this change ;)

custom_components/myskoda/coordinator.py
fursov: Sorry, could you clarify? I could not find the `last_updated` field anywhere in the repo.

WebSpider: I meant the LastUpdated(MySkodaSensor)

WebSpider (approved):
Lgtm

PR #368, opened 2024-11-29
Improve auxiliary heater to show up on all supported vehicles

prvakt:
Parking heater rewritten to support all scenarious and additionaly also `HVACMode.FAN_ONLY` mode is available if car supports `ACTIVE_VENTILATION`, otherwise only `HVACMode.HEAT`
- Request is created based on available capabilities and is different between PHEV, ICE and can contain following parameters:
--  `startMode` only if both `AUXILIARY_HEATING` and `ACTIVE_VENTILATION` are present (For cars where both heater and ventilation is available)
-- `targetTemperature` only if `AUXILIARY_HEATING_TEMPERATURE_SETTING` or `AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY`  present (PHEV and for example Kodiaq)
-- `durationInSeconds` only if `AUXILIARY_HEATING_TEMPERATURE_SETTING` not present
-- `heaterSource` only for `AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY` (PHEV)

Additionaly a `heater_duration` number entity is created for cars where targetTemperature can't be set (`AUXILIARY_HEATING` present, `AUXILIARY_HEATING_TEMPERATURE_SETTING` missing), So users can set their own value as this can't be set individualy like temperature

Would be great if we can deploy this PR as a alpha/beta version so users can test it first before it gets to the main 

Additionally because of #356, state change is not catched immediately in some cases because of parsing error of another service event


WebSpider:
Awesome contribution 🎉

Question: Do we need any changes in myskoda library or will 0.13.2 work?

custom_components/myskoda/climate.py
```diff
@@ -179,33 +190,116 @@ def __init__(self, coordinator: MySkodaDataUpdateCoordinator, vin: str) -> None:
...
+            ClimateEntityFeature.TURN_ON | ClimateEntityFeature.TURN_OFF
+        )
+
+        if self.vehicle.has_capability(
+            CapabilityId.AUXILIARY_HEATING_TEMPERATURE_SETTING
+        ) or self.vehicle.has_capability(
+            CapabilityId.AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY
+        ):
+            self._attr_supported_features |= ClimateEntityFeature.TARGET_TEMPERATURE
+
```
WebSpider, line 203: Have you tested this? I rolled back changes because defining `_attr_supported_features` didn't work for me in `__init__`

custom_components/myskoda/climate.py
```diff
@@ -260,13 +384,24 @@ async def async_turn_off(self):  # noqa: D102
...
+    def is_supported(self) -> bool:
+        """Return true if any supported capability is present."""
+        if self.vehicle.has_capability(
+            CapabilityId.AUXILIARY_HEATING
+        ) or self.vehicle.has_capability(
+            CapabilityId.AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY
+        ):
+            return True
+        else:
+            return False
```
WebSpider, line 407: This is too complex.

Try:

```python
def is_supported(self) -> bool:
    return self.vehicle.has_capability(CapabilityId.AUXLIARY_HEATING) or self.vehicle.has_capability(CapabilityID.AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY)
```

custom_components/myskoda/number.py
```diff
@@ -87,3 +88,60 @@ def required_capabilities(self) -> list[CapabilityId]:
 
     def forbidden_capabilities(self) -> list[CapabilityId]:
         return [CapabilityId.CHARGING_MQB]
+
+
+class HeaterDuration(MySkodaNumber, RestoreEntity):
```
WebSpider, line 93: Please rename to more specific `AuxiliaryHeaterDuration` across the board

custom_components/myskoda/coordinator.py
```diff
@@ -96,6 +96,7 @@ def __init__(
         self.update_air_conditioning = self._debounce(self._update_air_conditioning)
         self.update_vehicle = self._debounce(self._update_vehicle)
         self.update_positions = self._debounce(self._update_positions)
+        self.duration: float | None = None
```
WebSpider, line 99: Please store this in an appropriate section of `self.data` and update classes accordingly

custom_components/myskoda/climate.py
```diff
@@ -99,6 +100,16 @@ def hvac_action(self) -> HVACAction | None:  # noqa: D102
                 return HVACAction.COOLING
             return HVACAction.OFF
 
+    @property
+    def min_temp(self) -> float:
+        """Return the minimum temperature that can be set."""
+        return 16.0  # Restrict to a minimum of 16°C
```
WebSpider, line 106: On enyaq minimum is 15.5 (after 16.0 is goes to LO, after 29.5 it goes to HI in the app)

custom_components/myskoda/number.py
```diff
@@ -87,3 +88,60 @@ def required_capabilities(self) -> list[CapabilityId]:
...
+            self.coordinator.duration = self._attr_native_value
+
+        # Update HA state to reflect the restored value
+        self.async_write_ha_state()
+
+    def required_capabilities(self) -> list[CapabilityId]:
+        return [CapabilityId.AUXILIARY_HEATING]
+
+    def forbidden_capabilities(self) -> list[CapabilityId]:
+        return [CapabilityId.AUXILIARY_HEATING_TEMPERATURE_SETTING]
```
WebSpider, line 147: So, when `CapabilityId.AUXILIARY_HEATING_TEMPERATURE_SETTING` is present, this entity should not be created?

custom_components/myskoda/translations/en.json
```diff
@@ -121,6 +121,9 @@
         "number": {
             "charge_limit": {
                 "name": "Charge Limit"
+            },
+            "heater_duration": {
+                "name": "Heater Duration"
```
WebSpider, line 126: Please be more specific here as well. I used to think this was enough, only to find out that:

1) Changing a translation key after introduction is troublesome since it becomes part of the entity id
2) Entities will get added that resemble yours

prvakt:
> Awesome contribution 🎉
> 
> Question: Do we need any changes in myskoda library or will 0.13.3 work?

I'm confident to say that heater itself will definitely work with 0.13.3 with all cars (PHEV/ICE).. 
Not sure about the ventilation as there are 2 endpoints for starting the ventilation 
```
api/v2/air-conditioning/{vin}/active-ventilation/start
api/v2/air-conditioning/{vin}/auxiliary-heating/start (with startMode="VENTILATION")
```
my guess is that when you have auxiliary heater, you can use the second method as well which is already implemented so maybe it will work out of box :)

the first one is maybe for cars without auxiliary-heater but not sure if there is such a car available as I did not find a fixture with ACTIVE_VENTILATION without AUXILIARY_HEATER in our fixture directory.. But if there is such a car. Anyway I'm going to implement the first method into myskoda lib later and we can create a separate button for this case (only ventilation) as it doesn't make sense to use climate entity

custom_components/myskoda/climate.py
prvakt: yes, I have tested it and works as expected

So if there is no AUXILIARY_HEATING_TEMPERATURE_SETTING or AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY capability, climate entity will be created without target temperature feature (only on/off)

custom_components/myskoda/number.py
prvakt: yep, Heater Duration entity is created only for cars with AUXILIARY_HEATING but without AUXILIARY_HEATING_TEMPERATURE_SETTING..  Because if second is present, then heater is controlled using target temperature and timer is not present

WebSpider:
> > Question: Do we need any changes in myskoda library or will 0.13.3 work?
> 
> I'm confident to say that heater itself will definitely work with 0.13.3 with all cars (PHEV/ICE).. Not sure about the ventilation as there are 2 endpoints for starting the ventilation
> 
> my guess is that when you have auxiliary heater, you can use the second method as well which is already implemented so maybe it will work out of box :)
> 

Let's try this first then, and see if we need to expand in a follow-up PR.

custom_components/myskoda/coordinator.py
prvakt: created additional config class in self.data so we can store also other stuff if needed in future

prvakt:
fyi we should wait until #359 is finished because at least on PHEV there is always charging event fired before start-auxiliary-heating event, and because of the parsing error, AC state is never updated (just after regular update which can happen quite late depending on polling configuration).

WebSpider:
Please mark this as ready when you feel like we can merge this. Thanks!

custom_components/myskoda/icons.json
WebSpider, line 1: Please sync this file with `main`, there is no need to change tabspacing every PR

WebSpider (changes requested):
Thanks 👍

One minor remark about tabspacing in a json file, otherwise lgtm

custom_components/myskoda/icons.json
prvakt: I don't know what's going on with tabspacing.. Did not changed anything and I'm just following what all translations are using ( 4 spaces per tab ). When merging from main it always shows conflict in en.json...

custom_components/myskoda/climate.py
```diff
@@ -179,33 +190,116 @@ def __init__(self, coordinator: MySkodaDataUpdateCoordinator, vin: str) -> None:
...
+        ):
+            self._attr_supported_features |= ClimateEntityFeature.TARGET_TEMPERATURE
+
+    def _has_any_capability(self, cap: list[CapabilityId]) -> bool:
+        """Check if any capabilities in the list is supported."""
+        return any(self.vehicle.has_capability(capability) for capability in cap)
+
+    def _has_capability(self, cap: list[CapabilityId]) -> bool:
+        """Check if all capabilities in the list are supported."""
+        return all(self.vehicle.has_capability(capability) for capability in cap)
```
dvx76, line 210: nit: these could probably live in the base `MySkodaEntity` class since they could be useful for other entities too - can always move later though

custom_components/myskoda/climate.py
```diff
@@ -179,33 +190,116 @@ def __init__(self, coordinator: MySkodaDataUpdateCoordinator, vin: str) -> None:
...
+            [
+                CapabilityId.AUXILIARY_HEATING_TEMPERATURE_SETTING,
+                CapabilityId.AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY,
+            ]
+        ):
+            if ac := self._air_conditioning():
+                target_temperature = ac.target_temperature
+                if target_temperature is not None:
+                    return target_temperature
+        return None
```
dvx76, line 228: nit: `None` is the default return value when reaching the end of a method - no need to explicitly return it.

(same comment for other methods below)

WebSpider (approved):
Lgtm

custom_components/myskoda/coordinator.py
```diff
@@ -64,10 +64,16 @@ def __init__(
 Operations = OrderedDict[str, EventOperation]
 
 
+@dataclass
+class Config:
+    auxiliary_heater_duration: float | None = None
```
dvx76, line 69: Can you explain how this config works? I can see if unset in `self.data` we'll create a default `Config` object (which will have `auxiliary_heater_duration=None`), but where/how do set `auxiliary_heater_duration` to something else?

custom_components/myskoda/climate.py
```diff
@@ -217,39 +311,72 @@ def target_temperature(self) -> None | float:  # noqa: D102
...
+                        duration_in_seconds=self._duration_in_seconds,
+                        start_mode=AuxiliaryStartMode.VENTILATION,
+                    )
+                    _LOGGER.info("Starting ventilation [%s]", config)
+
+                    await self.coordinator.myskoda.start_auxiliary_heating(
+                        vin=self.vehicle.info.vin, spin=spin, config=config
+                    )
+                else:
+                    _LOGGER.error("Cannot start ventilation: No S-PIN set.")
```
dvx76, line 367: nit: general comment, you can often reduce indentation of a lot of code (making it more readable, with fewer lines getting broken up) by reversing the condition and returning early. Obviously only works if there's no more code to execute further down.

E.g. here

```suggestion
                if spin is None:
                    _LOGGER.error("Cannot start ventilation: No S-PIN set.")
                    return
                if (
                    ac.state != AirConditioningState.OFF
                    and ac.state != AirConditioningState.VENTILATION
                ):
                    _LOGGER.info("%s mode detected, stopping first.", ac.state)
                    await self.coordinator.myskoda.stop_air_conditioning(
                        self.vehicle.info.vin
                    )
                if ac.state == AirConditioningState.VENTILATION:
                    _LOGGER.info("%s already running.", ac.state)
                    return

                config = AuxiliaryConfig(
                    duration_in_seconds=self._duration_in_seconds,
                    start_mode=AuxiliaryStartMode.VENTILATION,
                )
                _LOGGER.info("Starting ventilation [%s]", config)

                await self.coordinator.myskoda.start_auxiliary_heating(
                    vin=self.vehicle.info.vin, spin=spin, config=config
                )
```

dvx76 (commented):
Just minor comments here and I am ok to merge as-is. I only skimmed over the code and to be honest I haven't been following all the aux-heater stuff in detail 😳 .

So if @WebSpider is good with this, I would suggest to merge it and consider my comments as future improvements.

I can try to run this code against my Enyaq to confirm it doesn't affect the climate entity.

Many thanks for all your work on the aux heater stuff. I'm definitely not familiar enough with it and it looks like there are a lot of corner cases and variations to handle.  💯

dvx76:
Found no problem uing the climate entity on Enyaq with this code 👍

custom_components/myskoda/coordinator.py
prvakt: heater duration is used for cars with `AUXILIARY_HEATING` but without `AUXILIARY_HEATING_TEMPERATURE_SETTING`.. Its used to defined for how long the auxiliary heater should be running before its automatically switched off. Because it is not possible to set it via own API method like target temperature I have created own (dummy) NumberEntity sensor which holds a value defined by user. This number is send in `start_auxiliary_heating`.. So the number is set from `AuxiliaryHeaterDuration` class (number.py) and used by `AuxiliaryHeater` (climate.py)

prvakt:
please review again, `async_set_hvac_mode` have been simplified and its more readable now..

WebSpider (approved):
lgtm

PR #371, opened 2024-11-29
New sensor: operation, report last and recent operation events

dvx76:
Add a new sensor, `operation`, which reports the state of the most recent event of type `EventType.OPERATION`. Attributes include additional details like operation name, error code, and a history of the previouys 4 operations.

Screenshot is from an early version which didn't filter the fields to be displayed in the history, but it gives an idea.

Bunch of formatting updates because json files had a mix of tabs and spaces 👀 

image


custom_components/myskoda/sensor.py
```diff
@@ -76,6 +78,53 @@ def _status(self) -> ChargingStatus | None:
                 return status
 
 
+class Operation(MySkodaSensor):
+    """Report the most recent operation."""
+
+    entity_description = SensorEntityDescription(
+        key="operation",
+        translation_key="operation",
+        device_class=SensorDeviceClass.ENUM,
```
WebSpider, line 87: Maybe this should be a diagnostic sensor, and maybe even hidden by default?

WebSpider (commented):
Great addition! :muscle: 

One small comment on the entity class we may want to use. 
Oh, and it turns out we have different default settings for jsons :rofl:

WebSpider (approved):
Otherwise lgtm!

custom_components/myskoda/sensor.py
dvx76: Diagnostic definitely - adding that now. Not sure about hidden. This could be useful for many users and it's harder to discover it if it's hidden by default :)

dvx76:
Rebased and added `entity_category=EntityCategory.DIAGNOSTIC`. Ready to merge. @WebSpider let me know if your comment can be resolved. Feel free to just merge.

WebSpider (approved):

PR #387, opened 2024-12-03
AuxiliaryHeater entity updated

prvakt:
logic changed in following way:
- if there is start/stop auxiliary heater event and no AUXILIARY_HEATING capability, update_air_conditioning is executed, otherwise update_auxiliary_heating
- AuxiliaryHeater entity is using data (state and targetTemp) either from `air_conditioning` or `auxiliary_heating` class based on the capabilities of the car


custom_components/myskoda/climate.py
```diff
@@ -205,19 +210,24 @@ def __init__(self, coordinator: MySkodaDataUpdateCoordinator, vin: str) -> None:
...
+            if ac := self._auxiliary_heating():
+                target_temperature = ac.target_temperature
+        elif self.has_any_capability(
+            [CapabilityId.AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY]
         ):
             if ac := self._air_conditioning():
                 target_temperature = ac.target_temperature
-                if target_temperature is not None:
-                    return target_temperature
+        return target_temperature
```
WebSpider, line 230: This sensor only `is_supported` when any of these capabilities are present:
```python
[
    CapabilityId.AUXILIARY_HEATING,
    CapabilityId.AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY,
]
```
Why do we care about the `None` case?

WebSpider:
Merge after #386 as it is based off of that source branch

WebSpider (changes requested):
Thanks for this quick update!

One remark about trying to cover a case that we have ruled out already.

WebSpider:
Please solve the merge conflict :+1:

prvakt:
> Please solve the merge conflict 👍

done

WebSpider (approved):
Simply lovely!

PR #389, opened 2024-12-03
Change default debounce behavior

WebSpider:
Many remarks about this integration include a "sluggish feel". 
The main reason for this, is that we use a so-called debouncer, to make sure that we do not poll the API to often, causing a lock-out (error 429)

Currently the default behaviour is:

-- Event happens
-- Debouncer is initiated and waits a default of 30s
-- Update is polled
-- Event happens
-- <etc>

This changes the default behaviour to:

-- Event happens
-- Debouncer is initiated and polls an update
-- All subsequent events of the same type are ignored for a default of 30s
-- Event happens
-- <etc>

This should make the integration feel less sluggish, at the risk of polling too early.

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -317,7 +317,7 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
             self.set_updated_vehicle(vehicle)
 
     def _debounce(
-        self, func: RefreshFunction, immediate: bool = False
+        self, func: RefreshFunction, immediate: bool = True
     ) -> RefreshFunction:
         return MySkodaDebouncer(self.hass, func, immediate).async_call
 


dvx76 (approved):
Great improvement. Never thought of this and had also wondered a few times why we're waiting before polling after an event (especially after operation completion events)

I do wonder if things will 'queue up' though, e.g.
1. MQTT event received
2. Immediate poll, debouncer starts 30s lock
3. Within 30s another event received
4. 30s pass, doe the poll from step 3 execute, or does another poll request need to be issued?

WebSpider:
> I do wonder if things will 'queue up' though, e.g.
> 
> 1. MQTT event received
> 2. Immediate poll, debouncer starts 30s lock
> 3. Within 30s another event received
> 4. 30s pass, doe the poll from step 3 execute, or does another poll request need to be issued?

The latter. Calls to the debouncer will get eaten for the period that cooldown is active.

PR #390, opened 2024-12-03
Disable MQTT connection during password check

WebSpider:
By default MQTT is connected during the config flow when we check if your password is valid.
This is unneeded and causes many listeners for MQTT changes, making mqtt messages get processed multiple times.

Not only is this unneeded and clutters the logfiles, it also can cause more API calls than needed and slow down HA

Also, we keep this extra connection around, with no clear purpose. So let's clean it up.


dvx76 (approved):
Good stuff!

PR #399, opened 2024-12-04
show HVACMode.FAN_ONLY also for AUXILIARY_HEATING_BASIC

prvakt:
From last few fixtures its seems like skoda changed `ACTIVE_VENTILATION` to `AUXILIARY_HEATING_BASIC`. Because of that ventilation was not available in AuxiliaryHeater entity. As I'm not sure if Skoda replaced that capability for all or only for some cars, I have modified the code to checks for both capabilities.


WebSpider (approved):
lgtm

PR #401, opened 2024-12-04
Fix OutsideTemperature sensor setup

pawelaugustyn:
Adds the OutsideTemperature (created in https://github.com/skodaconnect/homeassistant-myskoda/pull/388) as a HA sensor


WebSpider (approved):
Completely overlooked that. Thanks!

PR #407, opened 2024-12-04
update outsideTemperature sensor

prvakt:
outsideTemperature is either taken from AuxiliaryHeating response (if `AUXILIARY_HEATING` capability is available), otherwise from AirConditioning


WebSpider:
Please split this up in two PR's: 

One for the dependency of the library
One for the sensor update

WebSpider (changes requested):
Please do the lib dependency seperately

custom_components/myskoda/sensor.py
```diff
@@ -560,9 +560,17 @@ class OutsideTemperature(MySkodaSensor):
 
     @property
     def native_value(self) -> float | None:  # noqa: D102
-        if aux_heat := self.vehicle.auxiliary_heating:
-            if outside_temp := aux_heat.outside_temperature:
-                return outside_temp.temperature_value
+        aux_heat = (
+            self.vehicle.auxiliary_heating
+            if self.has_any_capability([CapabilityId.AUXILIARY_HEATING])
```
WebSpider, line 565: Why explicitly check one capability, but not the other? Either check both, or none. 

Preferably: Check none. The presence of the information alone in the `vehicle` tree should be enough to conclude we can use the information. The presence of the capability does not say anything conclusive about if the information is usable, so we need to check that regardless of the presence of the capability. 

These checks can simply be performed with the walrus ( := ) operator.

WebSpider (changes requested):
Thanks for this fix! :muscle: 

A few remarks about coding style and consistency with other parts of the code

WebSpider (approved):

WebSpider:
Thanks for all your hard work! :heart:

PR #416, opened 2024-12-07
Update LastUpdated sensor value on service events

fursov:
As discussed in https://github.com/skodaconnect/homeassistant-myskoda/pull/359, the car_captured_timestamp can be updated with timestamp value available from service events.

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -197,6 +197,9 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
 
     async def _on_charging_event(self, event: EventCharging):
         vehicle = self.data.vehicle
+        if vehicle.status and (timestamp := event.event.timestamp):
+            vehicle.status.car_captured_timestamp = timestamp
+
         data = event.event.data
 
         if vehicle.charging is None or vehicle.charging.status is None:
@@ -228,12 +231,18 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
         self.set_updated_vehicle(vehicle)
 
     async def _on_access_event(self, event: EventAccess):
+        if self.data.vehicle.status and (timestamp := event.event.timestamp):
+            self.data.vehicle.status.car_captured_timestamp = timestamp
         await self.update_vehicle()
 
     async def _on_air_conditioning_event(self, event: EventAirConditioning):
+        if self.data.vehicle.status and (timestamp := event.event.timestamp):
+            self.data.vehicle.status.car_captured_timestamp = timestamp
         await self.update_air_conditioning()
 
     async def _on_departure_event(self, event: EventDeparture):
+        if self.data.vehicle.status and (timestamp := event.event.timestamp):
+            self.data.vehicle.status.car_captured_timestamp = timestamp
         await self.update_positions()
 
     def _unsub_refresh(self):


WebSpider (changes requested):
Thanks for following up👍

Since there is no garuantee that a received event has a more recent timestamp than what we have in memory, can you add a check to see if the timestamp is more current than what we have in memory?

custom_components/myskoda/coordinator.py
```diff
@@ -228,12 +231,18 @@ async def _on_charging_event(self, event: EventCharging):
         self.set_updated_vehicle(vehicle)
 
     async def _on_access_event(self, event: EventAccess):
+        if self.data.vehicle.status and (timestamp := event.event.timestamp):
+            self.data.vehicle.status.car_captured_timestamp = timestamp
         await self.update_vehicle()
```
dvx76, line 236: Wouldn't `update_vehicle` already update this? For an access event a least the timestamp on the event isn't really representative of the `car_capture_timestamp`. That is, the event itself didn't really provide any updated data.

dvx76 (commented):
I guess this comes down to the question _what do we actually want to represent with `car_captured_timestamp`_ ?

Is it "the last time we got a full update of the car data" or is it "any data". Individual entities already report when they last changed value so I'm not sure if we really want/need a partial update (e.g. just a few charging entities after a charge event) to also update `car_captured_timestamp`. But I'm not against it either.

fursov:
> I guess this comes down to the question _what do we actually want to represent with `car_captured_timestamp`_ ?
> 
> Is it "the last time we got a full update of the car data" or is it "any data". Individual entities already report when they last changed value so I'm not sure if we really want/need a partial update (e.g. just a few charging entities after a charge event) to also update `car_captured_timestamp`. But I'm not against it either.

Yes, this comes down to getting same behavior as in the application or we will have own meaning for the sensor. As far as I see, if I go to the individual pages in the app, then I see different "updated" timestamp values there, meaning they display e.g. charging updated timestamp on the charging page, while the first overview page might have different time.
I also do not have strong opinion.
The best is to understand, what we use this info for. If it is informative only (e.g. no automations based on that) - I believe it does not really matter, we can have just most fresh timestamp there.

WebSpider:
> I guess this comes down to the question _what do we actually want to represent with `car_captured_timestamp`_ ?
> 
> Is it "the last time we got a full update of the car data" or is it "any data". Individual entities already report when they last changed value so I'm not sure if we really want/need a partial update (e.g. just a few charging entities after a charge event) to also update `car_captured_timestamp`. But I'm not against it either.

Looking at the responses we're seeing, there seems to be a difference in timestamps representing data obtained from the car and timestamps related to events.

Maybe it's a better idea to not use the same sensor for both events, but create a new sensor that represents the timestamp of latest event occurred. If someone then really wants to know 'when did anything happen', they can automate on any change of either sensor, or create their own template sensor that returns the most recent change.

dvx76:
> Maybe it's a better idea to not use the same sensor for both events, but create a new sensor that represents the timestamp of latest event occurred, as opposed to last poll occurred, which i guess resembles car_captured_timestamp the most.

I like this idea. We already have this for all events related to operations, through the `operation` sensor. I would prefer to not touch that one so that it can be used specifically in relation to operations and and their current/last state. But perhaps we can have a generate `event` sensor which just reports the last event received?

I would then leave `car_captured` as it is. It's a 1-1 mapping from a specific field returned by the API (`vehicle.status.car_captured_timestamp`).

For individual entities HA already reports the last change and last updated timestamps.

fursov:
Thanks for the feedback. That is fine to me.

Issue #370, opened 2024-11-29
Refactor coordinator._on_charging_event function


fursov:
Currently the function does following steps:
1. checks if `vehicle.charging` and `vehicle.charging.status` contain data
    - updates the status if it is not available, otherwise
    - updates the data to `vehicle.charging.status` with the data received from `change-soc` event
2. checks if `vehicle.driving_range` contains data
    - updates vehicle.driving_range if it is not available, otherwise
    - updates data to vehicle.driving_range from received change-soc event
3. updates vehicle using `set_updated_vehicle(vehicle)`

The flow is a bit suboptimal, partially because `update_charging()` and `update_driving_range()` implicitly call `set_updated_vehicle(vehicle)` inside so step 3 becomes redundant. Another drawback is having two places where event type should be analysed (in both step 1 and 2).

The flow should be following:
1. check if `vehicle.charging` and `vehicle.charging.status` contain data
    - if not, then call `update_charging()` and exit the function, because `update_charging()` will handle the vehicle update
2. check if `vehicle.driving_range` contains data
    - if not then call `update_driving_range()`, do not exit, proceed to 3
3. create match-case for the received event type
    - if it is `change-soc` - update data in all required structures and call `set_updated_vehicle(vehicle)`
    - if it is `charging-status-changed`, then call `update_charging()` since the event does not contain data, we need to have it updated explicitly

Other charging events can be added later if needed because currently I have no good understanding when those are fired.

WebSpider:
> Currently the function does following steps:
> 
> 1. checks if `vehicle.charging` and `vehicle.charging.status` contain data
>    
>    * updates the status if it is not available, otherwise
>    * updates the data to `vehicle.charging.status` with the data received from `change-soc` event
> 2. checks if `vehicle.driving_range` contains data
>    
>    * updates vehicle.driving_range if it is not available, otherwise
>    * updates data to vehicle.driving_range from received change-soc event
> 3. updates vehicle using `set_updated_vehicle(vehicle)`
> 
> The flow is a bit suboptimal, partially because `update_charging()` and `update_driving_range()` implicitly call `set_updated_vehicle(vehicle)` inside so step 3 becomes redundant. Another drawback is having two places where event type should be analysed (in both step 1 and 2).
> 
> The flow should be following:
> 
> 1. check if `vehicle.charging` and `vehicle.charging.status` contain data
>    
[...]

PR #417, opened 2024-12-07
Refactor _on_charging_event

fursov:
The refactoring is done according to #370 with minor modification. The flow is now following:

- if `vehicle.charging` is not available then it should be fetched
- if `vehicle.driving_range` is not available then it should be fetched
- the rest of the parsing is based on the fact which service event structure is in the event data. Currently there are only two options:
  - `ServiceEventChargingData` that contains some data that can be stored to the vehicle status and driving range structures
  - `ServiceEventData` that does not contain data. Since the event is received the status of the car has changed and it can be updated using `update_charging()` method

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -24,7 +24,10 @@ from myskoda.event import (
 )
 from myskoda.models.info import CapabilityId
 from myskoda.models.operation_request import OperationName, OperationStatus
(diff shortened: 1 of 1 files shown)


dvx76 (approved):
LGTM. Still feels like this event handling code needs more work but definitely an improvement.

Also don't love the use of `match/case`, and the resulting deep nesting of `match-case-if-if`, but that's not introduced by this PR and can be improved later.

PR #425, opened 2024-12-09
Improve incomplete response handling

WebSpider:
This improves incomplete response handling from MySkoda API by:

- Ignoring error 429 as well as 500, only logging an error for it
- Allowing the user endpoint to fail if we already have this information, since it does not update that often


prvakt (approved):
lgtm

prvakt (approved):

PR #430, opened 2024-12-11
Bump myskoda to 0.15.0

WebSpider:


prvakt:
@WebSpider should I try to quickly fix the primary/secondary range for CNG before this release or we will do it after getting fixtures?

prvakt (approved):
when I think about it, it will probably be better to adjust the driving range (primary/secondary) for CNG only after we get the fixtures so I can test everything first

Issue #342, opened 2024-11-23
Vehicle status render


prvakt:
Would be great if we can retrieve render of vehicle status as well.

Render needs to be retrieved trough MySkoda API with `HTTP GET` using full path from `get_status()` response -> /`renders`/`light_mode` or `dark_mode`/`one_x` or `one_and_half_x` or `two_x` or `three_x` depending on required size. threeX is the biggest one with size 888x1338px in my case

For example `HTTP GET` `https://mysmob.api.connect.skoda-auto.cz/api/v2/vehicle-status/render?carType=COMBI&vehicleState=1-1-3-1-0-0-0-0-0&lastModifiedAt=1730291332&dimension=3x&theme=LIGHT` -> where response is PNG image like the one below
![image](https://github.com/user-attachments/assets/388343ac-6a19-4123-b473-1589bc51dc0b)

WebSpider:
Ah, you did an actual request, that was on my todo list. Good to know it will show you the actual state as png!

WebSpider:
I've been tinkering with this one a bit. 
The main challenge is to obtain the image inside the clientsession of the library, that way we are authentication-proof.

PR #437, opened 2024-12-11
Add Image entities for vehicle status

WebSpider:
This adds 2 image entities that can display the vehicle status, as is done in the MySkoda App as well.
- One image is in 'Light' style
- One image is in 'Dark' style

The status is related to the window state, door state, bonnet state, etc.

Images are disabled by default, since it probably mostly is interesting to developers of custom cards.
They can be enabled manually in HA when going to the list of entities, and enabling them.

image


WebSpider:
Fixes #342

prvakt:
question 1: do we want to have those 2 under `SENSORS` because I think its more suitable for `DIAGNOSTICS`
question 2: I have noticed state for all images (also main render) is uknown.. I dont know if we want to handle state or not, but if yes maybe we can show something like available if image is present and unavailable if not?
```
    if is_available:
        self._state = "available"
    else:
        self._state = "unavailable"
    self.schedule_update_ha_state()
```
 
question 3: both images are same for my case, but from code I can see its retrieving correct path, so just want to check its same for you as well. (If it same, then does it make sense to have both?)
image
image

prvakt:
question 4: is it updated with the periodical vehicle update?

WebSpider:
> question 4: is it updated with the periodical vehicle update?

If there is an update to the state, the URL should change, triggering an update as part of the `ImageEntity`. Have not seen that happen yet, since my car is not close to my home at the moment.

prvakt:
1 ) asked ChatGPT about the sensor vs diagnostic section and it suggest also sensors :)
```
Suggestion for Your Case:
Given that the image likely represents the state of doors and windows in real-time, I recommend placing it under Sensors, as it primarily visualizes the output of door/window state sensors. If, however, it includes additional diagnostic elements like system errors or advanced analytics, then Diagnostics would be a better fit.
```

2 ) the state of all image entities can be implemented later once we agree what to show. 

4 ) I'm running this version for hour with poll interval set to 15 minutes.. Initially the state was correct, then I have opened door but state was never changed, even I see periodical update of all sensors have been triggered, Status Render image was not updated..

prvakt:
> The `ImageEntity` has some strange defaults. This should be corrected now.

Still doesnt work, I can see from logs that image url was changed from
```
https://mysmob.api.connect.skoda-auto.cz/api/v2/vehicle-status/render?carType=COMBI&vehicleState=1-1-1-1-0-0-0-0-0&lastModifiedAt=1733938917&dimension=3x&theme=LIGHT
```
to
```
https://mysmob.api.connect.skoda-auto.cz/api/v2/vehicle-status/render?carType=COMBI&vehicleState=1-1-3-1-0-0-0-0-0&lastModifiedAt=1733938917&dimension=3x&theme=LIGHT
```
but image was not updated and last updated timestamp was not changed either...

WebSpider:
Thanks, will continue in the mysteries of the `ImageEntity` tomorrow 😂

dvx76:
From https://developers.home-assistant.io/docs/core/entity/image/

>Implement async_image or image if your entity returns bytes of the image instead of providing a URL. Frontend will call async_image or image to fetch the image. If the image is fetched remotely, image data should be cached and the cache invalidated when image_last_updated is changed.

I would assume if the url is updated this would automatically invalidate the cache and force the url to be re-fetched. But maybe that's not happening?

custom_components/myskoda/image.py
```diff
@@ -59,6 +64,38 @@ def __init__(
         super().__init__(coordinator, vin)
 
 
+class StatusImage(MySkodaImage):
+    """A render of the current status of the vehicle."""
+
+    _attr_should_poll: bool = True
+
+    async def _fetch_url(self, url: str) -> httpx.Response | None:
```
prvakt, line 72: `_fetch_url()` is not called from anywhere.. is this maybe the problem why status image is never updated?

WebSpider: _fetch_url is used in the parent class

prvakt: ah OK.. fyi I have tried to put some more logs in image.py but both `_fetch_url()` and also `image_url()` are called only once and never after

WebSpider:
> From https://developers.home-assistant.io/docs/core/entity/image/
> 
> > Implement async_image or image if your entity returns bytes of the image instead of providing a URL. Frontend will call async_image or image to fetch the image. If the image is fetched remotely, image data should be cached and the cache invalidated when image_last_updated is changed.
> 
> I would assume if the url is updated this would automatically invalidate the cache and force the url to be re-fetched. But maybe that's not happening?

Rereading that piece of documentation, I think they mean we should implement a "last changed" property ourselves, so did that in last commit.

prvakt:
> Rereading that piece of documentation, I think they mean we should implement a "last changed" property ourselves, so did that in last commit.

We are still missing something. "Last changed" was updated correctly, but image was not changed. I can see in MySkoda app that image has changed and door is now opened, and also new binary sensors which I have created today were updated accordingly but image was not

WebSpider:
I will have a look at that parent class more thoroughly. This will not make v1.16.0.

custom_components/myskoda/image.py
WebSpider: I have added cache invalidation code, lets see if this solves it.

prvakt:
> I will have a look at that parent class more thoroughly. This will not make v1.16.0.

Good job! Can confirm it works as expected now ;)

prvakt (approved):

PR #447, opened 2024-12-13
Add seperate setup flow to coordinator

WebSpider:
This alters the coordinator setup as follows:

- When doing a data update (poll), we make a distinction between a regular update and one being made while the MySkoda integration is being setup
- During setup, we request only the minimal amount of data required to start. We do not fetch all sensor data anymore.
- During setup, we tell HA to schedule the MQTT connection for when we are done setting up
- During regular polling, we check if MQTT connection is still present, and try to reconnect if needed

This should allow the integration to succesfully set-up, even if MQTT is unavailable, and the REST API is being wonky instead of blocking HA startup or failing to initialise.


WebSpider:
Converted to draft since this probably needs a different approach

dvx76:
Why does this need a different approach?

I'm trying to understand how the coordinator is involved during the setup (ConfigFlow?) i.e. how does `_async_update_data` end up being called during the setup?

WebSpider:
> Why does this need a different approach?

The approach that is currently in this draft is mostly incomplete and has corner cases for multiple vehicles, integration reloading

> 
> I'm trying to understand how the coordinator is involved during the setup (ConfigFlow?) i.e. how does `_async_update_data` end up being called during the setup?

ConfigFlow is the configuration of the integration. Once the integration is configure, it can be setup. Those are two different things.

In setup, `async_setup` is called in `__init__.py`, which instantiates and inits the coordinator(s).

Have a look at the coordinators parent class, they are partially overriden by us only

WebSpider:
My idea is to only request data during first setup of coordinator that is needed for the actual initialisation, and then do a normal data pull.

This should hopefully make integration setup a lot more robust

WebSpider:
Fixes after running some tests while MQTT is down

custom_components/myskoda/coordinator.py
```diff
@@ -119,14 +121,77 @@ def __init__(
...
+                and self.data.vehicle.info.specification.model == "CitigoE iV"
+            ):
+                _LOGGER.debug(
+                    "Detected Citigo iV, requesting only partial update without health"
+                )
+                vehicle = await self.myskoda.get_partial_vehicle(
+                    self.vin,
+                    [
+                        CapabilityId.AIR_CONDITIONING,
+                        CapabilityId.AUXILIARY_HEATING,
```
prvakt, line 142: AUXILIARY_HEATING was removed in #491 from the partial update
```suggestion
```

prvakt (changes requested):
lgtm, just one minor change

WebSpider:
Good catch, will update the PR

PR #456, opened 2024-12-16
fix missing main render

prvakt:
render is taken from composite renders


custom_components/myskoda/device_tracker.py
```diff
@@ -84,8 +84,13 @@ def longitude(self) -> float | None:  # noqa: D102
     def extra_state_attributes(self) -> dict:
         """Return extra state attributes."""
         attributes = {}
-        if render := self.get_renders().get("main"):
-            attributes["entity_picture"] = render
+
+        render_list = self.get_composite_renders().get("unmodified_exterior_front")
+        if isinstance(render_list, list) and render_list:
+            for render in render_list:
+                if isinstance(render, dict) and "exterior_front" in render:
```
fursov, line 91: In the info return for my Suprb iV I see this:
```
                    {
                        "layers": [
                            {
                                "url": "https://skodavinimagecache.blob.core.windows.net/cache3v3s20201119/0F0FEKAcM-837-icSUg3f_sXOJRZCyGa-CK2vH0E59-4A6othXxnP8dOzuUaqJV-H9AsCqv16YbLX2EKOzn5fjwk-IaBMLRAHXUTFO-19201080dayvext_front1080.png",
                                "type": "REAL",
                                "order": 0,
                                "view_point": "exterior_front"
                            },
                            {
                                "url": "https://mspgwlivestorage.blob.core.windows.net/renders/cables/superb_3V_cable_charging_light_v1.png",
                                "type": "REAL",
                                "order": 1,
                                "view_point": "exterior_front"
                            }
                        ],
                        "view_type": "CHARGING_LIGHT"
                    },
```
So there can be list where there are two `exterior_front` entries and the second one is just a plug.
Of course, currently most probably it selects the one from `UNMODIFIED_EXTERIOR_FRONT` view type (where there is only one entry), but I think there should be some better selection for the URL than just `exterior_front` key.

custom_components/myskoda/device_tracker.py
```diff
@@ -84,8 +84,13 @@ def longitude(self) -> float | None:  # noqa: D102
     def extra_state_attributes(self) -> dict:
         """Return extra state attributes."""
         attributes = {}
-        if render := self.get_renders().get("main"):
-            attributes["entity_picture"] = render
+
```
fursov, line 87: would you keep `renders` as main source of the image and use `composite_renders` as backup?

custom_components/myskoda/device_tracker.py
prvakt: `render_list = self.get_composite_renders().get("unmodified_exterior_front")` ==> this selects UNMODIFIED_EXTERIOR_FRONT

 and this
```
if isinstance(render_list, list) and render_list:
            for render in render_list:
                if isinstance(render, dict) and "exterior_front" in render:
```
will go trough all urls and selects the one with "exterior_front" from UNMODIFIED_EXTERIOR_FRONT ;)

custom_components/myskoda/device_tracker.py
prvakt: yep this sounds like a good idea to keep main and use composite as fallback, I will change it

gschot: The composite_renders are useful as they represent the stacks of images in various situations. If you look at the data, you see there are various situations in the yaml: home, charging_light, charging_dark, etc. The list of exterior_side entries in that section represent the various (partial) images in the order they are stacked.

custom_components/myskoda/device_tracker.py
fursov: missed that one :)

custom_components/myskoda/device_tracker.py
prvakt: @gschot I know, but expecting that unmodified one will be the one where is nothing changed or stacked... this is just about changing entity picture for main render and device tracker as MySkoda api is sending now empty list for renders.. To have entity picture present in such a case, we select another picture from composite renders list.. (and my best guess is that unmodified will be always present and always single)

WebSpider: > yep this sounds like a good idea to keep main and use composite as fallback, I will change it

If you need to do a fallback, please log a debug message

gschot: @prvakt Got it! Thanks for the clarification.

custom_components/myskoda/image.py
```diff
@@ -70,7 +70,15 @@ class MainRenderImage(MySkodaImage):
...
-        return self.get_renders().get("main")
+        if render := self.get_renders().get("main"):
+            return render
+        else:
+            _LOGGER.debug("Main render not found, choosing composite render instead.")
+            render_list = self.get_composite_renders().get("unmodified_exterior_front")
+            if isinstance(render_list, list) and render_list:
+                for render in render_list:
+                    if isinstance(render, dict) and "exterior_front" in render:
+                        return render["exterior_front"]
```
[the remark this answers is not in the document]
prvakt: no need to return None as None is defined to be default return type
`def image_url(self) -> str | None:`

fursov (approved):
lgtm

WebSpider (approved):
lgtm

Issue #458, opened 2024-12-16
Citigo-e: mileage no longer available


pencoe:
Sensor "mileage" only delivers "unknown". 

Last working version: 1.14.2 (because the 1.15.x was never announced as an update i jumped to 1.15.3 from this)

Home Assistant 2024.12.3
MySkoda 1.15.3
Language: german

WebSpider:
We have made a change in v1.15 because the way we were requesting data was causing CitigoE to enter battery protection mode. This is an unintended side-effect from that change #431

pencoe:
will Mileage come back or is this someting we have to live with? 

Can't say how much it sucks that VAG (Skoda, SEAT, VW) does not offer a stable, documented API to their cars. I really appreciate that there are brave people like you that reverse engineer the communication between the App(s) and the appropriate server: an extra big THANKS to you and all contributors! ❤️

WebSpider:
It will be back in v1.16.1

WebSpider:
Please update to v1.16.1

PR #431, opened 2024-12-11
CitigoE iV partial update

WebSpider:
Hardcode a different update method for CitigoE as we learn how to cope with the limitations of this model.

This is expected to change over time, as we get more experience with this car model's behaviour. This model seems to dislike our polling of the vehicle health details very much, causing many errors in app and integration.

Fixes #212

PR #462, opened 2024-12-16
Introduce fallback for mileage sensor

WebSpider:
After disabling the `vehicle-health` endpoint for CitigoE in #431, the mileage sensor would become unpopulated, as reported in #458 and on Discord.

This introduces a fallback location for the mileage data.


prvakt (approved):
lgtm

PR #463, opened 2024-12-16
fix for fix missing main render

sonar98:
With pr  #456 the main image render is replaced with exterior_front if unavailable.
Problem is that exterior_front is not available for all vehicles.

If exterior_front is unavailable now exterior_side is used


WebSpider (changes requested):
Thanks for your help on this!

Please have a look at the failed tests, probably running `ruff format` on your repo will fix them :)

sonar98:
I think it is fixed now.

WebSpider (approved):
It works for now. This probably will be reformatted in the future to make it more readable, but this returns the function

PR #467, opened 2024-12-17
add Gas Range and Gas Level sensors

prvakt:
+ update Combustion Range and Fuel level to take value from engine with gasoline or diesel engine_type

### DIFF
--- a/custom_components/myskoda/icons.json
+++ b/custom_components/myskoda/icons.json
@@ -145,8 +145,14 @@
             "electric_range": {
                 "default": "mdi:ev-station"
             },
+            "gas_range": {
+                "default": "mdi:gas-station"
+            },
             "fuel_level": {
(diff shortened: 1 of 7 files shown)


WebSpider (changes requested):
There is no entity in this PR, only icons and translations

prvakt:
> Thanks for this great new feature! 🎅
> 
> Could you for future updates split out the new sensors from the changes to existing sensors, if they are not solving the same problem? Thanks! 🥂

This change is connected to GAS sensors actually... Before my change, fuel level and remaining combustion distance was taken always from primary engine and only for PHEV vehicles.. But HYBRID can be also gasoline+cng vehicle where primary engine is holding info about cng level and remaining distance - and because of that we need to take fuel level and combustion distance from the primary or secondary engine based on engine type

PR #490, opened 2024-12-23
Doc: template sensor for vehicle_in_motion

### LINKED ISSUE #471: Add Vehicle_Moving sensor like there is in the VW integration
I would like to be able to see whether or not the vehicle is currently moving. There is such a sensor in the VW integration, so my guess is that it should also be available for Skoda...

dvx76:
Fixes #471

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,15 @@ If your desired language is not available, please [open an issue](https://github
 
 ### Device Tracker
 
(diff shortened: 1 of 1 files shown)


WebSpider (approved):

sonar98:
there is a typo: achieve -> achieved

sonar98 (commented):
typo, achieve should be achieved

PR #491, opened 2024-12-24
remove AuxiliaryHeating from partial update for CitigoE IV

prvakt:
CitigoE IV doesnt have auxiliary heater as that's available only for ICE and PHEV vehicles

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -153,7 +153,6 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
                         self.vin,
                         [
                             CapabilityId.AIR_CONDITIONING,
-                            CapabilityId.AUXILIARY_HEATING,
                             CapabilityId.CHARGING,
                             CapabilityId.PARKING_POSITION,
                             CapabilityId.STATE,


WebSpider (approved):

prvakt:
> CitigoE IV doesnt have auxiliary heater as that's available only for ICE and PHEV vehicles

now I'm wondering why auxiliary-heating data was requested if car doesn't have this capability and also see from log that departure timer was requested as well.. looks like some Citigoes are not fulfilling this condition or something else is wrong
```
if (
  self.data.vehicle.info.device_platform == "MBB"
  and self.data.vehicle.info.specification.model == "CitigoE iV"
):
```

prvakt:
> > CitigoE IV doesnt have auxiliary heater as that's available only for ICE and PHEV vehicles
> 
> now I'm wondering why auxiliary-heating data was requested if car doesn't have this capability and also see from log that departure timer was requested as well.. looks like some Citigoes are not fulfilling this condition or something else is wrong
> 
> ```
> if (
>   self.data.vehicle.info.device_platform == "MBB"
>   and self.data.vehicle.info.specification.model == "CitigoE iV"
> ):
> ```

Ahh I see, it requests the full data for the first time and then if Citigo is detected, it request partial data next time.. then still wondering why auxiliary-heating was requested if not in capability...  

Switching to draft and will request fixtures from that car first..

PR #498, opened 2024-12-30
Add Departure Timer 1,2,3 switch

prvakt:
This will add a switch for all 3 departure timers. Additionally whole timer configuration is stored in custom attributes

### DIFF
--- a/custom_components/myskoda/switch.py
+++ b/custom_components/myskoda/switch.py
@@ -2,7 +2,7 @@
 
 import logging
 from datetime import timedelta
-from typing import Any, Optional
+from typing import Any
 
 from homeassistant.components.switch import (
     SwitchDeviceClass,
@@ -511,7 +511,7 @@ class DepartureTimerSwitch(MySkodaSwitch):
         super().__init__(coordinator, vin)  # Initialize parent class (MySkodaEntity)
         self.timer_id = timer_id  # Store the specific timer ID for each subclass
 
-    def get_timer(self) -> Optional[DepartureTimer]:
+    def get_timer(self) -> DepartureTimer | None:
         """Retrieve the specific departure timer by ID."""
         if departure_info := self.vehicle.departure_info:
             if departure_info.timers:
@@ -523,14 +523,12 @@ class DepartureTimerSwitch(MySkodaSwitch):
                     ),
                     None,
                 )
-        return None
 
     @property
     def is_on(self) -> bool | None:
         """Check if the timer is enabled."""
         if timer := self.get_timer():
             return timer.enabled
-        return None
 
     @property
     def extra_state_attributes(self) -> dict[str, Any]:
--- a/custom_components/myskoda/switch.py
+++ b/custom_components/myskoda/switch.py
@@ -2,6 +2,7 @@
 
 import logging
 from datetime import timedelta
+from typing import Any, Optional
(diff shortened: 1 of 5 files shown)


custom_components/myskoda/switch.py
```diff
@@ -496,3 +501,113 @@ async def async_turn_on(self, **kwargs):  # noqa: D102
...
+                    None,
+                )
+        return None
+
+    @property
+    def is_on(self) -> bool | None:
+        """Check if the timer is enabled."""
+        if timer := self.get_timer():
+            return timer.enabled
+        return None
```
WebSpider, line 533: This line can be removed. `None` is returned by default

custom_components/myskoda/switch.py
```diff
@@ -496,3 +501,113 @@ async def async_turn_on(self, **kwargs):  # noqa: D102
...
+
+class DepartureTimerSwitch(MySkodaSwitch):
+    """Base class for departure timers, handling common functionality."""
+
+    def __init__(self, coordinator, vin, timer_id: int, **kwargs):
+        """Initialize the departure timer switch."""
+        super().__init__(coordinator, vin)  # Initialize parent class (MySkodaEntity)
+        self.timer_id = timer_id  # Store the specific timer ID for each subclass
+
+    def get_timer(self) -> Optional[DepartureTimer]:
```
WebSpider, line 514: While this is valid, we have multiple points in the code where we use `State | None` instead of `Optional[] `.  Please use `State | None` for consistency sake.

custom_components/myskoda/switch.py
```diff
@@ -496,3 +501,113 @@ async def async_turn_on(self, **kwargs):  # noqa: D102
...
+            if departure_info.timers:
+                return next(
+                    (
+                        timer
+                        for timer in departure_info.timers
+                        if timer.id == self.timer_id
+                    ),
+                    None,
+                )
+        return None
```
WebSpider, line 526: This would then be the default as well, and can be omitted

custom_components/myskoda/switch.py
```diff
@@ -496,3 +501,113 @@ async def async_turn_on(self, **kwargs):  # noqa: D102
...
+
+    @property
+    def is_on(self) -> bool | None:
+        """Check if the timer is enabled."""
+        if timer := self.get_timer():
+            return timer.enabled
+        return None
+
+    @property
+    def extra_state_attributes(self) -> dict[str, Any]:
```
WebSpider, line 536: Storing things in `extra_state_attributes` bring some side effect. For instance, this is never picked up by the `recorder` and this does not have any history.

Are the `extra_state_attributes` the right place for this, or could we maybe use a `schedule` helper?

WebSpider (commented):
You did it 👍😎

Thanks a lot!

A few remarks around code consistency, otherwise a great addition, thanks!

custom_components/myskoda/switch.py
prvakt: yeah I know, just wanted to put it somewhere so users can actually see what is inside the timer they are enabling.. Its only for information purposes. In future I'm planning to implement also a service call so users can actually configure their own timer from our integration but not today 😄

WebSpider (approved):
Lgtm

PR #500, opened 2024-12-30
add generate diagnostics feature for the config entry section

prvakt:
this will add possibility to generate diagnostics directly from the integration section for all available vehicles. Format and structure is same as for device diagnostics.


prvakt:
image

this entry was available but generating empty file which caused confusion for some users.. With this PR, it will be possible to download diagnostics from integration page (for all available vehicles) and also for a specific vehicle from device page

WebSpider (approved):
Awesome 👍

Issue #494, opened 2024-12-27
Enyaq NO_MOD1_4_VEHICLES error - Could not obtain GDC token


fm4991:
since a few days my homeassistant said myskoda unavailable. Reload didn't help.
So i deleted the Hub for my car and tried to set it up again.

But i get an Unkown Error during setup.
the only thing i find in the logs is:

![image](https://github.com/user-attachments/assets/5139664e-bb8c-4d80-99ca-dea738fc114d)

prvakt:
please update to latest version, seems like you are running older version where MQTT was failing because of change made by MySkoda - this was fixed in `1.15.3` and latest version is `1.17.2`


just fyi this problem is still not completely fixed - seems like MySkoda implemented some kind of restriction for the UUID and some UUIDs are working and some not.. so can happen that you are good to go during first connect, sometimes second or third but sometimes it fails multiple times in row which may result in increased reconnect backoff to a minute or two..   

**_NOTE:_** we are using random uuid for connection to MQTT as this was done by myskoda app as well, but probably they have some range of allowed UUIDs - until we will not find how they build this UUID, we can just generate a random one and retry with another one if it fails

fm4991:
Hey @prvakt,

thanks for the fast answer. Yeah seems like i did a manual install and never updated it 🙈🫣

but now i have another problem.
I add my Account this seems to be successfull.
But there are no entities created.

prvakt:
> Hey @prvakt,
> 
> thanks for the fast answer. Yeah seems like i did a manual install and never updated it 🙈🫣
> 
> but now i have another problem. I add my Account this seems to be successfull. But there are no entities created.

If you have a CitigoE iV, then please use `1.16.1`. There is a problem with one endpoint in `1.17.x` which doesn't work for Citigo and we are fixing this problem right now. 

If you have another car, maybe you are locked out from MySkoda right now. When creating issue next time, please put some more details at least about your car model and version used (ideally with diagnostics report from MySkoda integration) - we will be able to advise you better

fm4991:
I have a Skoda Enyaq, was using latest 1.17.2 but also tried 1.16.1 now here are the debugging logs from the try with 1.16.1:
[home-assistant_myskoda_2024-12-30T13-08-00.286Z.log](https://github.com/user-attachments/files/18274846/home-assistant_myskoda_2024-12-30T13-08-00.286Z.log)

[...]

PR #502, opened 2024-12-30
(chore) Config updates: Cache list of seen vins in ConfigEntry

WebSpider:
This PR is to be merged after #528 and adds:

- V2.2 adds the list of last seen vins. In case the garage is empty for some reason, we still have the VINs and continue initialization

Main use for this is to be able to start up even if the garage endpoint gives us an error.


WebSpider:
This should help to prevent ending up in situations such as #494

WebSpider:
Needs a bit more testing still, conversion does not seem to always happen

dvx76:
If I understand correctly this will be closed in favor of #528 + future PRs?

WebSpider:
No, this has been rewritten to contain only the 'vinlist' config data, that enables a cache of last seen vins

prvakt (approved):
difficult to test this functionality, but code looks OK, lets try it 👍

WebSpider:
> difficult to test this functionality, but code looks OK, lets try it 👍

I've tested it on my install to see if the migration works, and it does. the `ConfigEntry` (stored in `/config/.storage/core.config_entries`) is updated

Issue #450, opened 2024-12-15
Add last event sensor


fursov:
Per discussion in https://github.com/skodaconnect/homeassistant-myskoda/pull/416
Sensor that would be similar to last operation sensor and contain information about latest received event.

PR #506, opened 2025-01-01
Implement service event sensor with history

fursov:
This is almost a copy from operations history sensor. Implements issue #450.

The sensor is in diagnostics section and displays the latest event timestamp. I decided to use timestamp instead of the event name because if event with same name arrives then, if I understand right, there will be no sensor state change fired in HA.


custom_components/myskoda/coordinator.py
```diff
@@ -196,6 +206,7 @@ async def _on_mqtt_event(self, event: Event) -> None:
                 await self._on_air_conditioning_event(event)
             if event.topic == ServiceEventTopic.DEPARTURE:
                 await self._on_departure_event(event)
+            self.async_set_updated_data(self.data)
```
dvx76, line 209: Doesn't this already end up being called indirectly by the various `_on_xxxx_event` methods above? I _think_ we need to avoid calling this multiple times?

custom_components/myskoda/coordinator.py
```diff
@@ -247,6 +258,11 @@ async def _on_operation_event(self, event: EventOperation) -> None:
         ]:
             await self.update_departure_info()
 
+    def _update_service_event_history(self, event: ServiceEvent):
+        # Store the last MAX_STORED_SERVICE_EVENTS events
+        self.service_events.append(event)
+        self.service_events = self.service_events[-MAX_STORED_SERVICE_EVENTS:]
```
dvx76, line 264: Python can do this natively with https://docs.python.org/3/library/collections.html#collections.deque

Instead of initializing as a list you would initialize as `deque([], MAX_STORED_SERVICE_EVENTS)` and then you can just blindly append and it will only ever have the last `MAX_STORED_SERVICE_EVENTS` items. Probably even append with `appendleft` then when iterating over the list you'll get the events from newest to oldest.

fursov: Good hint, I'll take a look on deque

custom_components/myskoda/coordinator.py
fursov: In this commit I removed all calls to `async_set_updated_data` from `_on_xxxx_event` functions (`set_updated_vehicle` called it). But now I think this is wrong, because they are all called from debouncer and if the debouncer delays and schedules the call for later, then the vehicle data will not be updated after rest api call returns.
I will bring `async_set_updated_data` back to `set_updated_vehicle`

dvx76: I'm not clear if there is still potential for calling `async_set_updated_data` multiple times (within a single event being processed). But I don't know the implications if it does end up being called multiple times. Obviously ideally it's only called one time after all the 'data' has been collected/updated.

WebSpider: > I'm not clear if there is still potential for calling `async_set_updated_data` multiple times (within a single event being processed). But I don't know the implications if it does end up being called multiple times. Obviously ideally it's only called one time after all the 'data' has been collected/updated.

Nothing big will happen, as long as they all contain the complete vehicle state we want.

dvx76 (approved):
LGTM, thanks for the updates

PR #511, opened 2025-01-02
(chore) Reference ConfigEntry consistently

WebSpider:
This renames references to ConfigEntry from `config` to `entry`:
- `config` is already used in relation to configurations of various climate entries
- The naming was inconsistent, sometimes `config`, sometimes `entry`

Found this while working on #447 and it annoyed me :grin:

### DIFF
--- a/custom_components/myskoda/__init__.py
+++ b/custom_components/myskoda/__init__.py
@@ -48,11 +48,11 @@ PLATFORMS: list[Platform] = [
 ]
 
 
-async def async_setup_entry(hass: HomeAssistant, config: ConfigEntry) -> bool:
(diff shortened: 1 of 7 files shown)


prvakt (commented):
question - is it not better to use `config_entry` instead `entry`/`config` ? It's unlikely to clash with other variables in future since it’s specific and from name its already clear that this object represents a ConfigEntry

WebSpider:
I chose `entry` because its in the HA example docs for integration development, but if we think `config_entry` is better, sure 👍

prvakt:
> I chose `entry` because its in the HA example docs for integration development, but if we think `config_entry` is better, sure 👍

I don't have a preference, so I'll leave the decision up to you :)

prvakt (approved):

PR #517, opened 2025-01-04
workaround for unavailable sensors

prvakt:
Affected sensors: `AddBlueRange`, `CombustionRange`, `GasRange`, `GasLevel`, `Range`

After last update, they become unavailable because they are created only when some specific data from driving_range enpoint is present. This was changed as this data is not requested during initialization sequence

Those sensors will be created for vehicles based on the required or forbidden capabilities. Side effect is that some will be created also for vehicles which doesn't have this feature (for example AddBlueRange for all combustion vehicles). Because of that a `available()` property is used to make the sensor unavailable/available based on the data 

So for example users with combustion vehicles will see something like this:
image
Sensor is created, but set to unavailable. They can hide the sensor manually as its not possible to hide/unhide sensors from code

### DIFF
--- a/custom_components/myskoda/sensor.py
+++ b/custom_components/myskoda/sensor.py
@@ -247,10 +247,18 @@ class AddBlueRange(MySkodaSensor):
             if range.ad_blue_range is not None:
                 return range.ad_blue_range
 
-    def is_supported(self) -> bool:
+    @property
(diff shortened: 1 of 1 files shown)


WebSpider (approved):
Thanks for the quick fix!

PR #520, opened 2025-01-05
Pull data from REST API when HA has launched

WebSpider:
After PR #447 all the sensors do not have a value after booting HA. This fixes that.

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -174,6 +174,7 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
                     "MySkoda has finished starting up. Scheduling post-start tasks for vin %s.",
                     vin,
                 )
+                await self._async_update_data()
                 try:
                     coord = hass.data[DOMAIN][config.entry_id][COORDINATORS][vin]
                     if not coord.myskoda.mqtt and not coord._mqtt_connecting:


custom_components/myskoda/coordinator.py
```diff
@@ -174,6 +174,7 @@ def _async_finish_startup(hass, config, vin) -> None:
                     "MySkoda has finished starting up. Scheduling post-start tasks for vin %s.",
                     vin,
                 )
+                await self._async_update_data()
```
prvakt, line 177: we need different approach because `await` can only be used with asynchronous functions

I have tried to fix it using code below, and from logs I saw data has been pulled from all REST endpoints, but sensors still remained unknown until first scheduled polling
```
            async def _finish_startup(hass: HomeAssistant) -> None:
                """Tasks to execute when we have finished starting up."""
                _LOGGER.debug(
                    "MySkoda has finished starting up. Scheduling post-start tasks for vin %s.",
                    self.vin,
                )
                # Schedule the async update and MQTT connection
                await self._async_update_data()
                try:
                    coord = hass.data[DOMAIN][self.entry.entry_id][COORDINATORS][
                        self.vin
                    ]
                    if not coord.myskoda.mqtt and not coord._mqtt_connecting:
                        hass.async_create_task(coord._mqtt_connect())
                except KeyError:
                    _LOGGER.debug("Could not connect to MQTT. Waiting for regular poll")

            # Schedule `_finish_startup` to run after startup
            async_at_started(hass=self.hass, at_start_cb=_finish_startup)
            return State(vehicle, user, config, operations)
```

prvakt (changes requested):

custom_components/myskoda/coordinator.py
dvx76: I wonder if we want to just revert back to pulling everything during initialization (leaving just the MQTT connection to be deferred later). I understand we want to avoid failing the init when just one of the API endpoints is acting up, but maybe this current approach is introducing more issues/complexity than it is solving?

WebSpider: I have been testing this as well, and my idea from last night is indeed flawed.

We probably need redo this by making a difference between the setup, first fetch and regular fetch. HA has options to do this, and there are plenty of examples present, but it turns out 5am is not my most bright moment 😆

Will work on this today some more.

PR #523, opened 2025-01-05
fix sunroof sensor not being created

prvakt:
missed this one in #517 - Sunroof sensors is created for all vehicles, and set unavailable for those without this feature


Schafy:
Since updating to 1.17.2 today, sunroof and Electric range entities are marked unavailble - both were working before. Model is an Enyaq.

prvakt:
> Since updating to 1.17.2 today, sunroof and Electric range entities are marked unavailble - both were working before. Model is an Enyaq.

please wait until new version is released

Schafy:
Was just meant as info - all fine

PR #524, opened 2025-01-05
Revert partial update when in SETUP state

WebSpider:
Using partial update in its current form leaves sensors without value between startup and first poll.
Disable the feature for now until we solve this.

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -161,7 +161,9 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
             _LOGGER.debug("Performing initial data fetch for vin %s", self.vin)
             try:
                 user = await self.myskoda.get_user()
-                vehicle = await self._async_get_minimal_data()
+                # - Disabled until we find a way to schedule an update when HA is started
+                # vehicle = await self._async_get_minimal_data()
+                vehicle = await self._async_get_vehicle_data()
             except ClientResponseError as err:
                 handle_aiohttp_error(
                     "setup user and vehicle", err, self.hass, self.entry


prvakt:
now sensors are available right after boot, but there is still this error message in logs:
```
2025-01-05 16:26:25.059 DEBUG (MainThread) [custom_components.myskoda.coordinator] MySkoda has finished starting up. Scheduling post-start tasks for vin TMBxxxxxxxxxxxxxx.
2025-01-05 16:26:25.059 DEBUG (MainThread) [custom_components.myskoda.coordinator] Connecting to MQTT.
2025-01-05 16:26:25.060 DEBUG (MainThread) [custom_components.myskoda.coordinator] Finished fetching myskoda data in 11.648 seconds (success: True)
2025-01-05 16:26:25.061 DEBUG (MainThread) [custom_components.myskoda.device_tracker] Main render not found, choosing composite render instead.
2025-01-05 16:26:25.065 ERROR (MainThread) [homeassistant] Error doing job: Future exception was never retrieved (None)
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
TypeError: 'NoneType' object is not callable
2025-01-05 16:26:25.238 DEBUG (MainThread) [myskoda.myskoda] Trace: GET https://mysmob.api.connect.skoda-auto.cz/api/v1/users - response: 200 (596 bytes) {...}
```

additionally it seems like `initial data fetch` is triggered 2 times one after another.. if you want to see the logs, send me a message in discord

WebSpider:
Yes, it has always been triggered a few times:

- once in `__init__.py` due to https://github.com/skodaconnect/homeassistant-myskoda/blob/a194f16cdc7142038aecc1d3171e413507509179/custom_components/myskoda/__init__.py#L90
- once more just the login in `__init__.py` due to https://github.com/skodaconnect/homeassistant-myskoda/blob/a194f16cdc7142038aecc1d3171e413507509179/custom_components/myskoda/__init__.py#L64
- Once more that probably is related to the `DataUpdateCoordinator` class, but i havent found the details yet

The `NoneType` i havent traces down yet, this PR was only to repopulate the entities.

Issue #497, opened 2024-12-29
Is it possible to create the sensors with names by default in English?


h3llrais3r:
It seems that my sensors are being created with Dutch names when adding the integration.
Is it using it from my local settings, or is it using the language from my skoda profile to determine the language?
Is it possible to force them in English instead?
I don't like a mix of the sensor names in different languages.
I rather use translations for the friendly names, not for the real sensor names.

h3llrais3r:
It seems that if I remove the `nl.json` file from the translations folder, it doesn't create the sensors in Dutch, but in English.
Not sure this is the best way?
Should I put back the translations file after that the sensors are created?

korttoma:
I was also surprised to see that all entities have Swedish names for me. I agree with @h3llrais3r on this. I don't like sensor names mixed in different languages, English names is to be preferred. 
Makes it easier also if I want to report a bug for a particular sensor if the name of the sensors have the same name for everyone.

WebSpider:
This is a side-effect from making the integration fully translated from the start that we have not looked into.
Personally, I like the feature, but think the default should be English, with local names being configurable. 

I wouldn't mind a PR that makes that a reality.

WebSpider:
#502 makes some preparations for this

WebSpider:
Looking at the HA documentation, changing this is possible, but a bit tedious. I will have a go at it somewhere this month.

WebSpider:
I've looked into this and asked some questions to the HomeAssistant developers.

Short: This is expected behaviour from HA, and considered a feature by HA. An integration cannot change this.
If you want the entity id's to be created in English, you should set the HA system language to English, and the UI language of your useraccount to your native language.

Settings > System > General changes the system language (which is what the entity ids will be created in by default)
Profile > General changes the UI language (which is what the entity names will be created in for you)

h3llrais3r:
> Settings > System > General changes the system language (which is what the entity ids will be created in by default)
> Profile > General changes the UI language (which is what the entity names will be created in for you)

If this is working, it would be great. For now both are put in Dutch, so that might be the reason.
Going to try this out and keep you posted.

h3llrais3r:
@WebSpider Thanks for this great tip, as it works as expected. (I didn't know it was working like this)

To conclude: 
- If you want all your sensors to be created by default in English, you have to put the system language to 'en'.
  `Settings > System > General changes the system language`
[...]

Issue #496, opened 2024-12-29
Remove "in KM" from "Next inspection" and "Oil service" sensors


cholek3:
Hi,

First of all, thank you for your work on this add-on.

Do API requests also provide imperial measurements for sensors? I have my Kamiq set for UK market, so all measurements are in miles, but add-on provides ```sensor.skoda_kamiq_range```, ```sensor.skoda_kamiq_next_inspection_in_km```, ```sensor.skoda_kamiq_oil_service_in_km``` and ```sensor.skoda_kamiq_mileage``` are in kilometres. If so, could we have those sensors providing measurements in miles, please?

I am having difficulties converting kilometres into miles, as I am getting different factors:
My current measurements from addon and what it shows in car:
~~range: 249m (car) / 402km (addon) = 1.614457831325301~~
~~mileage: 117m (car)/117.2m (trip mileage) /188km (addon) = 1.606837606837607/1.604095563139932~~
127.6m / 205.4km = 1.609717868338558

cholek3:
Well, I guess I will answer myself to this feature request. Just downloaded diagnostics file, and I see API only reports those values in kilometres, so I had to create template sensor:
```
- platform: template
  sensors:
    skoda_kamiq_range_in_miles:
      friendly_name: "Skoda Kamiq Range in miles"
      unit_of_measurement: 'm'
      device_class: distance
      icon_template: mdi:gas-station
      value_template: "{{ ((states('sensor.skoda_kamiq_range') | float(0) / 1.609717868338558 | float(0) )) | round(0) }}"

    skoda_kamiq_next_inspection_in_miles:
      friendly_name: "Skoda Kamiq service in miles"
      unit_of_measurement: 'm'
      device_class: distance
      icon_template: mdi:car-wrench
      value_template: "{{ ((states('sensor.skoda_kamiq_next_inspection_in_km') | float(0) / 1.609717868338558 | float(0) )) | round(0) }}"

    skoda_kamiq_mileage_in_miles:
      friendly_name: "Skoda Kamiq mileage in miles"
      unit_of_measurement: 'm'
      device_class: distance
      icon_template: mdi:counter
      value_template: "{{ ((states('sensor.skoda_kamiq_mileage') | float(0) / 1.609717868338558 | float(0) )) | round(0) }}"

    skoda_kamiq_oil_service_in_miles:
      friendly_name: "Skoda Kamiq oil service in miles"
      unit_of_measurement: 'm'
      device_class: distance
[...]

PR #528, opened 2025-01-05
(chore) Config migration flow

WebSpider:
This PR adds 2 things:
- Config migration flow: With config migration flow we can migrate our config and values between versions. It now becomes possible to rename recalculate values, rename entities, change option defaults, etc within a version upgrade
- Add `unique_id` to config entries: The config entries we created, which are used at integration setup, did not have a unique identifier. This is now added, and set to the user ID generated for your MySkoda account. Unique ID's are used within HA widely, but in this case it is particularly useful to support reauthentication (if you changed your password)

This is a partial reimplementation of #502, as that PR made too many changes at once and a preparation for #496 and #497


custom_components/myskoda/__init__.py
```diff
@@ -112,3 +120,67 @@ async def _async_update_listener(hass: HomeAssistant, entry: ConfigEntry):
...
+            try:
+                myskoda = myskoda_instantiate(hass, entry, mqtt_enabled=False)
+                await myskoda.connect(entry.data["email"], entry.data["password"])
+                user = await myskoda.get_user()
+                unique_id = user.id
+            except AuthorizationFailedError as exc:
+                raise ConfigEntryAuthFailed("Log in failed for %s: %s", DOMAIN, exc)
+            except Exception as exc:
+                _LOGGER.exception("Login with %s failed: %s", DOMAIN, exc)
+                return False
```
dvx76, line 165: question: what's the difference in behavior (by HA) when this returns `False` vs when an exception is raised? `False` ignores the migration and let's the init complete and an exception causes the init to fail?

dvx76 (approved):
Cool. Thanks for figuring this out! Code looks good to me.

custom_components/myskoda/__init__.py
WebSpider: `False` means the migration did not succeed, but lets the integration complete setup. When raising `ConfigEntryAuthFailed`, HA signals that there is a problem with authentication, and either starts a `ConfigFlow` of type `reauth`, which we dont have yet, so it falls back to an error in the UI.

PR #534, opened 2025-01-07
Remove battery deviceclass from TargetBatteryPercentage

### LINKED ISSUE #518: Battery level in "Devices Tab" showing charging limit
First of all: thanks a lot, seems to work already great with my Enyaq (before facelift)

I see the following issue: 
grafik

In the last row the "70%" is not the current battery level (is 59 at the moment), but the configured charging limit set. 

Clearly not an real problem, but should be the current level, I guess. At least the "old" skodaconnet instance showed that instead. 

BR, Florian

WebSpider:
Fixes #518


prvakt (approved):
lgtm

PR #538, opened 2025-01-08
Bump myskoda to 0.17.0

WebSpider:
- Adds airconditioning timers
- Avoids getting health info if this would wake up the car
- Dependency updates


prvakt (approved):
Lgtm

PR #562, opened 2025-01-13
Bump myskoda to 0.17.2

WebSpider:


prvakt:
we have changed window_heating_state to be default = None so WindowHeatingSwitch must be updated

```
    def is_on(self) -> bool | None:  # noqa: D102
        if ac := self.vehicle.air_conditioning:
            if window_heating_state := ac.window_heating_state:
                return (
                    window_heating_state.front == OnOffState.ON
                    or window_heating_state.rear == OnOffState.ON
                )
```

prvakt (approved):
LGTM

PR #567, opened 2025-01-14
change required capabilities for some sensors

prvakt:
ChargingPower, ChargingRate, TargetBatteryPercentage are now created if `EXTENDED_CHARGING_SETTINGS` and `CHARGING` capability is present


WebSpider:
Looming at the fixtures we have, this indeed solves the issue for Kodiaq iV. I don't think we have a recent Superb iV on file, so I guess we will find out if it works by releasing this :+1:

WebSpider (approved):
lgtm

PR #576, opened 2025-01-15
removed translated states from translation files

### LINKED ISSUE #384: remove sensor state values from translation
The myskoda integration is using translation for sensor values, which are also translated by default within HomeAssistant.
Example: Device_Tracker -> "home" == "Anwesend" (de.json) or several Binary_sensors.

Therefore, the translation is in some areas not consistent with the translation provided by HomeAssistant itself for such values.

The Integration should only provide translated values, when HomeAssistant does not provide a valid / usefull translation for a sensor, whenever possible, the translation should use the default translation provided by HomeAssistant.

ChristophCaina:
by removing the states from the translation files, HomeAssistant should use its default translation for binary_states according to their device_class.

Only, if a language is not available in HomeAssistant itself, the state should be default in english.

fix #384


ChristophCaina:
json error will be fixed... sorry ^^

WebSpider:
Have you tested this, and if so, on what version of HA and in which language?

Last time i tried this, it simply did english for all untranslated states.

ChristophCaina:
I did some tests with single sensors, before I made the changes.
I have now replaced all translation files with the ones I've provided in the PR...

After rebooting, everything looks as expected and is using the default language set in HA

DE
image

FR
image

did you maybe had the state translation still in your english json file?
HA will always use the translation provided by the integration.
If a specific file does not have a translation, but the english one does have, it will use the english translation from the integration.

Only, if there's no translation given at all, it will use its own translations.

ChristophCaina:
> Thanks for testing this, and for sharing your findings. 🙏
> 
> I will re-test this myself to see if we can for instance use a per-entity approach, where most are default from HomeAssistant, and some are translated.

this should already be given in the current translation files.
I've only removed the translations for binary_sensors and device_tracker.

Anything else that will have multiple states will be done with these translation files - since those are not translated by HA per default.

custom_components/myskoda/translations/en.json
```diff
@@ -156,12 +84,7 @@
         },
         "device_tracker": {
             "device_tracker": {
-                "name": "Position",
-                "state": {
-                    "home": "Home",
-                    "not_home": "Away",
-                    "vehicle_in_motion": "Motion detected"
-                }
```
WebSpider, line 164: We specifically have this feature for detecting the vehicle is in motion and have this state translated. Please don't remove this one

WebSpider (changes requested):
I have tested this, and indeed works as intended, except for the device tracker.

Can you fix the merge conflicts as well?

custom_components/myskoda/translations/en.json
ChristophCaina: Hi,
All right - I will have a look into this as well.

The problem here is, that at least the state "home" and "away" should be supported by HomeAssistant.
I will check, if it is working with having only the state "vehicle_in_motion" as a translation.

But it will take some time, until I can do this

WebSpider: I already checked, and it doesn't work if you leave them out. So for this sensor i'd like to propose we make an exception and keep `home` and `not_home` translations in the integration.

ChristophCaina: that's unfortune, since this sensor was initially the one I've started this topic :D
The translated state was different from the state HA was using by default...

I know, this might be a 'breaking' change - but according to the documentation of HA, there are basically two types of device_tracker available.

1: It is providing the GPS position - in this case, it can have any gps coordinate as a value
2: it is using a "router" (like fritzbox tools integration) - then it should provide only two states - home / not_home - similar to a binary sensor.

I would suggest to implement the device_tracker in a way that is consistent with the HA standard.
I guess, it does provide GPS coordinates, right?

In this case, the information if the vehicle is moving or not could be an additional binary sensor.
This would also allow it, to use this entity / sensor specificly in conditions for automations:

example:
If vehicle is away & moving
etc.

ChristophCaina: thanks for the feedback.
I'll revert the changes for the device tracker within the next days.

dvx76:
Is this PR stale?

ChristophCaina:
sorry, didn't had the time to do the changes due to some other private things ongoing the last couple of weeks :-(

ChristophCaina:
I should be able to finish this this weekend ... sorry for the delay

WebSpider:
Please let us know if we can do anything to drive this forward 👍

ChristophCaina:
I haven't put a high priority on this atm - as it is nothing that improves the integration as it is.
therefore, sorry - had a lot of other high priority topics on a private level the last months :-(

let me check - I should find some spare time this weekend and I'll set a reminder for me :)

WebSpider:
Superseded by #800. Thanks @ChristophCaina for your effort!

PR #596, opened 2025-01-29
Car Info entity sensor

pawelaugustyn:
A similar sensor was present in the old integration and I missed it a lot, so I decided to make up my own. Maybe someone else will also find it useful.


WebSpider (approved):
lgtm

PR #600, opened 2025-01-30
Improve config flow error

WebSpider:
- Improve config flow error translatables
- Add detection for Terms and Conditions and Marketing updates

This improves user experience when MySkoda is having issues, or if you need to relogin for profile updates


prvakt (approved):
👍

PR #607, opened 2025-02-03
Bump myskoda to 0.19.0

WebSpider:
The library adds:
- Possibility to add timestamps in most used models
- Bugfixes regarding the User model
- Dependency updates


prvakt (approved):
🆗

Issue #608, opened 2025-02-04
Periodic updates are deferred when charging is in progress


fursov:
I would like to report one misbehaviour. The scenario is following:
- in my setup the integration has polling period of 10 minutes
- the car is Superb iV 2020, quite old model. It reports charging events always **without data**, so those basically are informative that something has happened and if we need to fetch the additional info, then we have to call the update manually.

I noticed, that at some point I started to get regular "Unknown" state for the battery percentage, this mainly happens when the car in in charging state:
![Image](https://github.com/user-attachments/assets/c6ea5a29-e54d-4a64-b1d3-1f4914fe788c)

I checked the logs. So what happens is that while charging the car sends regular updates about charging by the event `charging-status-changed`, the event comes in about 9 minutes (less than polling interval). On this event the integration calls function `self.set_updated_vehicle(vehicle)` which is visible in the logs as `Manually updated myskoda data`. And it seems, that after this the periodic update is not called because the HA considers that vehicle received the manual update and periodic poll is not needed.

So why then the state becomes "Unknown". This has started to happen after this update https://github.com/skodaconnect/myskoda/pull/338. All charging events are now coming as `ServiceEventWithChargingData` where missing fields are `none` so for my car this means the charging state is `none` that turns to "Unknown" on battery sensors.

I attach the logs from this night charging (periodic update responses payload are removed to simplify anonymizing, they are not relevant for the issue). The updates are coming regularly until `2025-02-04 04:02:00`, when script starts charging. After that the server sends `charging-status-changed` regularly and there is no periodic updates anymore until `2025-02-04 06:15:49` just because when charge level is closer to 100%, the charging event updates are coming with slower rate from the car (longer than 9 minutes).
I also inserted few debug prints for myself that start with `CHARGING DEBUG`.

So, I think, there are couple of issues:
- the https://github.com/skodaconnect/myskoda/pull/338 should have a followup/fix to continue sending events without data as ServiceEventData, not as ServiceEventChargingData
- it is not clear for me, what should be done to enable periodic updates even if we call `self.set_updated_vehicle(vehicle)` because even if it is called when we receive e.g. charging info update with data, it does not mean we updated all other sensors.

Please, share your opinion.

[charging.log](https://github.com/user-attachments/files/18654286/charging.log)

WebSpider:
[...]

PR #610, opened 2025-02-04
fix unknown states from _on_charging_event

prvakt:
There were couple of places inside `_on_charging_event()` where state was assigned without checking if its exists in the event data which caused sensor to go to `unknown` state in some cases as described in #608


WebSpider (approved):
Thanks! 👍

PR #614, opened 2025-02-05
Implement user endpoint cache of 6 hours

WebSpider:
The data in the user endpoint rarely changes, yet we request it on every update.
This implements an internal cache of the user data, skipping the update if the timestamp is still valid.


custom_components/myskoda/coordinator.py
```diff
@@ -205,7 +206,20 @@ async def _async_finish_startup(hass: HomeAssistant) -> None:
 
         # Obtain user data. This is allowed to fail if we already have this in state.
         try:
-            user = await self.myskoda.get_user()
+            if not self.data.user or not self.data.user.timestamp:
+                user = await self.myskoda.get_user()
+            else:
+                if (
+                    datetime.now(UTC) - timedelta(hours=CACHE_USER_ENDPOINT_IN_HOURS)
+                    < self.data.user.timestamp
```
dvx76, line 214: Am I crazy or should this be `datetime.now(UTC) - timedelta(hours=CACHE_USER_ENDPOINT_IN_HOURS)
                    > self.data.user.timestamp` ?

Suggestion: might be clearer to rewrite like this:

```python
            if self.data.user and self.data.user.timestamp:
                cache_expiry_time = self.data.user.timestamp + timedelta(hours=CACHE_USER_ENDPOINT_IN_HOURS)
                if datetime.now(UTC) > cache_expiry_time:
                    _LOGGER.debug(
                        "Updating user - cache expired at %s", self.data.user.timestamp
                    )
                    user = await self.myskoda.get_user()
                else:
                    _LOGGER.debug("Skipping user update - cache is still valid.")
            else:
                user = await self.myskoda.get_user()
```

dvx76 (approved):

PR #621, opened 2025-02-12
Add support for odometer update event

WebSpider:
The odometer service-event was added in `myskoda v0.20.0`, this adds handling of this event in the integration.

If the event is fired, the maintance info will be updated, and if there is health information, this will also be updated and pushed into HA.

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -20,6 +20,7 @@ from myskoda.event import (
     EventAccess,
     EventAirConditioning,
     EventDeparture,
+    EventOdometer,
     EventOperation,
     ServiceEvent,
     ServiceEventTopic,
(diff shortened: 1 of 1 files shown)


custom_components/myskoda/coordinator.py
```diff
@@ -359,13 +362,39 @@ async def _on_air_conditioning_event(self, event: EventAirConditioning):
...
         self.data.vehicle = vehicle
         self.async_set_updated_data(self.data)
 
+    async def update_odometer(self) -> None:
+        _LOGGER.debug("Updating odometer information for %s", self.vin)
+
+        maint_info = health_info = vehicle = None
+
+        try:
+            maint_info = await self.myskoda.get_maintenance(self.vin)
```
dvx76, line 381: Does this need debouncing? I don't know how often we can expect these odo events.

custom_components/myskoda/coordinator.py
```diff
@@ -359,13 +362,39 @@ async def _on_air_conditioning_event(self, event: EventAirConditioning):
...
         return
 
     def set_updated_vehicle(self, vehicle: Vehicle) -> None:
         self.data.vehicle = vehicle
         self.async_set_updated_data(self.data)
 
+    async def update_odometer(self) -> None:
+        _LOGGER.debug("Updating odometer information for %s", self.vin)
+
+        maint_info = health_info = vehicle = None
```
dvx76, line 378: nit: `vehicle` doesn't need initializing

WebSpider: Without it the linter tripped, so i added it

custom_components/myskoda/coordinator.py
WebSpider: Good call, we best add a debouncer.

dvx76 (approved):

PR #624, opened 2025-02-13
Implement reauthentication flow

### LINKED ISSUE #622: Unable to update password after changing it in the MySkoda app
**Describe the bug**
After changing the password in the MySkoda app, the integration for Home Assistant does not provide an option to update the password. This results in an authentication error, and the only workaround is to remove and reinstall the integration. However, this leads to the loss of entities and historical data in Home Assistant.

**Is this working in MySkoda app?**
Yes, I was able to log out and log back in successfully in the MySkoda app (latest version on Android). The issue is only with the Home Assistant integration.

**Latest working release**
The issue is not specific to a release but occurs when changing the password in the MySkoda app.

**Debug logs**
Currently, I don't have debug logs enabled, but I can provide them if needed. Let me know the necessary steps to enable and collect them.

**Installation:**

Home Assistant variant: Hass OS - 2025.2.3

**Additional context**
A feature to update the password directly within the integration settings or prompt the user for new credentials when authentication fails would be helpful. Currently, the only solution involves deleting and reinstalling the integration, which is not user-friendly and results in the loss of all entity configurations and historical data.

WebSpider:
- Implement reauthentication flow
- Replace named config variables with constants

Reauthentication flow enables users to change their password when logging in fails.
Fixes #622


dvx76 (approved):
LGTM assuming it was tested :)

WebSpider:
> LGTM assuming it was tested :)

Could have tested this at a better moment :angel: 
Some improvements added.

dvx76 (approved):

PR #657, opened 2025-03-09
Remove unit from inspection entities

### LINKED ISSUE #496: Remove "in KM" from "Next inspection" and "Oil service" sensors

dvx76:
Entity IDs and names should not include units (in this case km) since the unit could be converted by homeassistant (e.g. miles) in which case the entity name becomes confusing.

Both are derived from the translations.

Use the same name for the inspection in days and in distance, which was already the case for at least one translation. Change the icon for the inspection in distance (matching the icon for mileage) to more easily differentiate them.

Do not modify the entity (translation-)keys to keep them in line with the underlying model/API responses.

Fixes #496


WebSpider (approved):
Lgtm

PR #659, opened 2025-03-10
Update charging logic to only update when engine is electric

### LINKED ISSUE #655: Incorrect Estimated Fuel Range on Skoda Superb PHEV
**Describe the bug**
I have a Skoda Superb PHEV, and the estimated fuel range is consistently incorrect. However, when I manually reload the entities, the correct value appears.
I've noticed that this issue always occurs while charging the car, even though charging should not affect the estimated distance on fossil fuel.

Image

Yes

**Latest working release**
unknown

**diagnostics**

[myskoda-01JFJ68RWEE52Q2AYDTP5KN7F6-Škoda Superb Combi-3a621c3e2ddeb61a3339b4145200964c.json](https://github.com/user-attachments/files/19125319/myskoda-01JFJ68RWEE52Q2AYDTP5KN7F6-Skoda.Superb.Combi-3a621c3e2ddeb61a3339b4145200964c.json)

**Installation:**
 - Home Assistant variant: Hass OS
 - Python library versions: see diagnostics

WebSpider:
Previously, any charging update received via MQTT would result in an update of the `primary_engine` always.
For Hybrid vehicles this may result in incorrect sensor updates, since the primary engine can be a combustion engine.

This PR fixes this by only updating charging information if the engine is an electric one.

Fixes #655

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -357,7 +357,10 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
                 status.state = event_data.state
         if vehicle.driving_range:
             per = vehicle.driving_range.primary_engine_range
-            ser = vehicle.driving_range.secondary_engine_range
+            ser = False
+
+            if vehicle.driving_range.secondary_engine_range:
+                ser = vehicle.driving_range.secondary_engine_range
 
             if event_data.soc:
                 if per.engine_type == EngineType.ELECTRIC:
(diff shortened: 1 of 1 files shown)


custom_components/myskoda/coordinator.py
```diff
@@ -355,14 +356,24 @@ async def _on_charging_event(self, event: EventCharging):
             if event_data.state:
                 status.state = event_data.state
         if vehicle.driving_range:
+            per = vehicle.driving_range.primary_engine_range
+            ser = vehicle.driving_range.secondary_engine_range
```
dvx76, line 360: At least `secondary_engine_range` can be `None` which would result in an `AttributeError` here.

dvx76 (approved):

PR #661, opened 2025-03-10
Remove fixtures button

WebSpider:
Has been deprecated earlier. Let's remove this now as we no longer use this.

A note should be made in release notes that we did this, as someone may use automations to press this button.


dvx76:
Does this need a migration step to actually delete the button entity in existing installations?

WebSpider:
If we want to remove it from the existing entities, yes. Right now it will be abandoned, waiting for the user to clean it up.
I will add some code to delete the entity by it's `unique_id`

PR #665, opened 2025-03-10
Force data refresh when an operation is done

### LINKED ISSUE #542: Lock state not represented lock correctly
**Describe the bug**
When I close the lock of my car using the lock entity, I can see the lock get locked in the MySkoda app (and physically of course). The state of the lock in the MySkoda app is unlocked (as is correct). The switch for the lock entity in HA goes down, but the wording (state) remains "locked". When switching the switch back up to locked the car actually gets locked (in MySkoda app and physically). All this time the state of the lock entity remains "locked", as well for the states of the binary_sensors related to the lock. 

Yes it is. 

**Latest working release**
Just noticed this now (latest release) as I never used the lock before through this integration.

**Debug logs**

**Installation:**
 - Home Assistant variant: Hass OS
 - Python library versions:
awake      1.0
packaging  24.2
pathspec   0.12.1
pip        24.3.1
pulsemixer 1.5.1
pyparsing  3.1.4
PyYAML     6.0.2
setuptools 70.3.0
yamllint   1.35.1

**Additional context**
Has this to do with the polling interval. I set it to 1 minute before writing this report, but then it didn't work. Now it seems it started to work...?
skoda

WebSpider:
This effectively disables the use of the debouncer when an Operation type event reports a status has changed.

Rationale here: We know something has changed, and we know the car is online at the moment, so wakeup is not needed.
Using a debouncer in this case has proved some updates to be skipped completely.

Fixes #542

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -306,12 +306,12 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
             OperationName.STOP_WINDOW_HEATING,
             OperationName.SET_AIR_CONDITIONING_TIMERS,
         ]:
-            await self.update_air_conditioning()
+            await self._update_air_conditioning()
         if event.operation.operation in [
             OperationName.START_AUXILIARY_HEATING,
             OperationName.STOP_AUXILIARY_HEATING,
         ]:
-            await self.update_auxiliary_heating()
+            await self._update_auxiliary_heating()
         if event.operation.operation in [
             OperationName.UPDATE_CHARGE_LIMIT,
             OperationName.UPDATE_CARE_MODE,
@@ -320,16 +320,16 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
             OperationName.STOP_CHARGING,
             OperationName.UPDATE_AUTO_UNLOCK_PLUG,
         ]:
-            await self.update_charging()
+            await self._update_charging()
         if event.operation.operation in [
             OperationName.LOCK,
             OperationName.UNLOCK,
         ]:
-            await self.update_status(immediate=True)
+            await self._update_status()
         if event.operation.operation in [
             OperationName.UPDATE_DEPARTURE_TIMERS,
         ]:
-            await self.update_departure_info()
+            await self._update_departure_info()
 
     async def _on_charging_event(self, event: EventCharging):
         vehicle = self.data.vehicle


dvx76:
The debouncer should never result in an update being skipped completely right?

I'm not convinced about this change. Do we not get the same result, while still spacing out API requests, if we use `immediate=True` everywhere? The first function call gets executed right away and _if_ there is another call it will still be executed but after 30s.

For #542 I wonder if perhaps the problem is that the state in the API itself isn't updated immediately. E.g. we start a LOCK operation, then when we get an operation completion event we refresh data from the API but the API still thinks the vehicle is unlocked?

dvx76:
> If you use `immediate=True`, then any subsequent update in the debounce timer is eaten, from the way I read [home-assistant/core@`ad126a7`/homeassistant/helpers/debounce.py#L24](https://github.com/home-assistant/core/blob/ad126a745a59dbc124e61a2585196339ee2157ba/homeassistant/helpers/debounce.py#L24)

That's not my understanding. I admit I haven't studied the Debouncer code in detail but I ran the following test to confirm my understanding:

```python
async def debounce_me():
    _LOGGER.info(f"debounce_me function requested at {datetime.now(tz=UTC)}")

# Triggering from a Button for convenience:
    async def async_press(self) -> None:
        debouncer = Debouncer(
            self.hass, _LOGGER, cooldown=10, immediate=True, function=debounce_me
        )
        _LOGGER.info("Debounce 1")
        await debouncer.async_call()
        await asyncio.sleep(2)
        _LOGGER.info("Debounce 2")
        await debouncer.async_call()
```

Output:

```
2025-03-12 09:51:12.888 INFO (MainThread) [custom_components.myskoda.button] Debounce 1
2025-03-12 09:51:12.888 INFO (MainThread) [custom_components.myskoda.button] debounce_me function requested at 2025-03-12 08:51:12.888756+00:00
2025-03-12 09:51:14.889 INFO (MainThread) [custom_components.myskoda.button] Debounce 2
2025-03-12 09:51:22.890 INFO (MainThread) [custom_components.myskoda.button] debounce_me function requested at 2025-03-12 08:51:22.890341+00:00
```

So the first `debouncer.async_call()` results in the debounced function running immediately , as expected. The second `debouncer.async_call()` call is debounced and the function runs 10 seconds after the first execution.

WebSpider:
> For #542 I wonder if perhaps the problem is that the state in the API itself isn't updated immediately. E.g. we start a LOCK operation, then when we get an operation completion event we refresh data from the API but the API still thinks the vehicle is unlocked?

Interesting. Since this works properly in the app, maybe the app assumes a lock state change upon receipt of the event?
I wonder how they then would know the lock is closed or opened.

WebSpider:
I remembered we discussed something similar in skodaconnect, and here it is:

https://github.com/skodaconnect/homeassistant-skodaconnect/issues/248

PR #672, opened 2025-03-12
Update all relevant sensors when receiving EventCharging

### LINKED ISSUE #484: SKoda Enyaq. Power and Range are not updated
I have noticed that the addon does not update the Charging Power and the Range while the car is being charged.
This bug occurs sometimes not always.

The old addon skoda connect does not show this issue

WebSpider:
Always update `charging` and `driving_range` when we receive an event of type `EventCharging`.

The data contained in the charging event itself does not have all the related sensor information, causing charging releated sensors not to be updated at all when `EventCharging` is emitted more frequently than the polling interval.

Fixes #484

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -335,12 +335,16 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
 
     async def _on_charging_event(self, event: EventCharging):
         vehicle = self.data.vehicle
-        update_charging_request_sent = False
+        _charging_updated = False
+        _range_updated = False
+
+        # See if we have relevant data
         if vehicle.charging is None or vehicle.charging.status is None:
             await self.update_charging()
-            update_charging_request_sent = True
+            _charging_updated = True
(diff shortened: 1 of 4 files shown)


custom_components/myskoda/coordinator.py
```diff
@@ -378,16 +382,14 @@ async def _on_charging_event(self, event: EventCharging):
...
+        # Make sure we update relevant data since the event has incomplete data
+        if not _charging_updated:
             await self.update_charging()
-        else:
-            self.set_updated_vehicle(vehicle)
+
+        if not _range_updated:
+            await self.update_driving_range()
+
+        self.set_updated_vehicle(vehicle)
```
dvx76, line 392: I'm wondering if there is still a point in pulling data out of the charging event and into `vehicle.charging` and `vehicle.driving_range` if we end up updating both here anyway (which will overwrite both objects with whatever comes back from the API). The normal flow will be that we already have a version of both objects at the start of this method.

If we're going to update from the API anyway, then I think it might be better to just do it at the start either way (so without any `is None` conditions), and then update specific field with what we received in the event.

PR #695, opened 2025-03-14
Swap source for mileage sensor, health endpoint is now cached

### LINKED ISSUE #693: Odometer strange values fluctuating
Logger: homeassistant.components.sensor.recorder
Bron: components/sensor/recorder.py:317
integratie: Sensor (documentatie, problemen)
Eerst voorgekomen: 14:30:10 (1 gebeurtenissen)
Laatst gelogd: 14:30:10

Entity sensor.skoda_enyaq_kilometerstand from integration myskoda has state class total_increasing, but its state is not strictly increasing. Triggered by state 48432 (48434.0) with last_updated set to 2025-03-14T13:26:43.819569+00:00. Please create a bug report at https://github.com/skodaconnect/homeassistant-myskoda/issues

Home Assistant variant: Hass OS
Version: 2025.3.2

Myskoda version: 1.21.2

Image

Image

WebSpider:
In release v1.21.2 caching was added for the health endpoint to avoid 429 errors on this healthpoint for some car models. This sensor used this endpoint, which made the results being cached.

Swap this for the mileage report that we get from another endpoint, which seems to have a much higher tolerance for 429 errors for all models.

Fixes #693

### DIFF
--- a/custom_components/myskoda/sensor.py
+++ b/custom_components/myskoda/sensor.py
@@ -532,11 +532,11 @@ class Mileage(MySkodaSensor):
 
     @property
     def native_value(self) -> int | None:  # noqa: D102
-        if health := self.vehicle.health:
-            return health.mileage_in_km
-        # If we have disabled the health endpoint, use this as fallback
-        elif maint_report := self.vehicle.maintenance.maintenance_report:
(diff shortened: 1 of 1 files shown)


dvx76 (approved):
Good catch! 🙌

PR #702, opened 2025-03-17
Switch to YAML issue templates

dvx76:
Deduplicate issue templates and switch to YAML templates. Also use issue types instead of labels.

Ref: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

Examples in my fork (using labels instead of issue type because those are only available in organizations 😭 ): https://github.com/dvx76/homeassistant-myskoda/issues/2 (click New Issue to see the forms and links)


dvx76:
> I'm all for this!
> 
> Any idea if the stale-bot needs adjustments too?

Good point. Looking at https://github.com/actions/stale it doesn't look like it supports the issue type yet. It should still work as-is based on the `stale`/`no-stale` labels. Do we want to add the `no-stale` label by default for all feature issues?

I also added a `triage` label by default, so we can easily see which new issues haven't been reviewed yet. The idea is that we then remove that label after reviewing it.

WebSpider:
> Do we want to add the `no-stale` label by default for all feature issues?

:+1:

> I also added a `triage` label by default, so we can easily see which new issues haven't been reviewed yet. The idea is that we then remove that label after reviewing it.

Makes perfect sense! :+1:

dvx76:
Added a link to search for issues (open _and_ closed) and default `no-stale` label for feature requests

WebSpider (approved):
lgtm

PR #707, opened 2025-03-18
Fix AC timer availability

### LINKED ISSUE #706: Air-conditioning timers are unavailable
### Please confirm:

HAOS 2025.3.3

### Car model

Enyaq 80x

### Latest working release

1.21.0 or 1.21.1 (do not remember)

### Bug Description

After updating to version 1.21.3 the Air-conditioning timers are now always unavailable. 

### Debug Logs

```bash

```

WebSpider:
Fixes #706 

In #662 UI improvements to pushing buttons, toggling switches etc wee introduced, making these entities available only when no other action is currently running.

This introduced a regression due to the fact that AC data only is available after startup, since we improved startup times by only requesting initially what's needed for entity creation.


dvx76 (approved):
👍

PR #716, opened 2025-03-24
Implement sensor indicating time left until AC has reached target

### LINKED ISSUE #678: Missing sensors Skoda Enyaq
For the Skoda Enyaq we had some more sensors, are you able to add these to the MySkoda app as well?

Is the vehicle moving or not: binary_sensor.skoda_enyaq_sportline_vehicle_moving
Pre-heat time left: sensor.skoda_enyaq_sportline_climatisation_time_left

Adjust charging limit: 

alias: Instelling Laadlevel
description: ""
mode: restart
triggers:
  - entity_id:
      - input_number.laadlevel
    trigger: state
conditions: []
actions:
  - data:
      device_id: xxx
      limit: "{{ states('input_number.laadlevel')|int }}"
    action: skodaconnect.set_charge_limit

WebSpider:
This introduces a sensor for vehicles that advertise having Air Conditioning:

`estimated_time_left_to_reach_target_temperature`

The value is calculated from the API `estimated_date_time_to_reach_target_temperature`, which indicates a point in time that is estimated when the AC reaches set time.

A time difference between now and that point in time is calculated. 
When that time difference is negative, the sensor will indicate 0, to make it clear there is no time left to reach target temperature.

Fixes #678

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ If your desired language is not available, please [open an issue](https://github
 - Fuel Level
 - Combustion range
 - Electric range
+- Estimated Time To Reach Target Temperature
 
 #### Last Updated sensor
 
--- a/custom_components/myskoda/sensor.py
+++ b/custom_components/myskoda/sensor.py
@@ -1,6 +1,6 @@
 """Sensors for the MySkoda integration."""
 
-from datetime import datetime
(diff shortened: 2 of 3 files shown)


dvx76 (approved):
Nice!

PR #720, opened 2025-03-24
Set version to v1.21.6

WebSpider:


WebSpider:
I'll bump to 1.22 later.

PR #725, opened 2025-03-26
Replace config field names with their constants

WebSpider:
Small cleanup. This bothered me reading through the code.
Also, add proper unloading of the MQTT open connection upon unload.


dvx76 (approved):
Nice!

PR #726, opened 2025-03-26
Simplify event handling in coordinator

WebSpider:
An attempt to simplify mqtt event handling in the coordinator.
By implementing event handler maps, we should be able to expand for more future events easily.

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -4,7 +4,7 @@ from collections import OrderedDict, deque
 from collections.abc import Coroutine
 from dataclasses import dataclass
 from datetime import datetime, timedelta, UTC
-from typing import Callable
+from typing import Any, Awaitable, Callable, Dict, cast
 
 from aiohttp import ClientError
 from aiohttp.client_exceptions import ClientResponseError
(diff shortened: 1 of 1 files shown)


custom_components/myskoda/coordinator.py
```diff
@@ -271,16 +271,19 @@ async def _on_mqtt_event(self, event: Event) -> None:
...
-                await self._on_charging_event(event)
-            if event.topic == ServiceEventTopic.ACCESS:
-                await self._on_access_event(event)
-            if event.topic == ServiceEventTopic.AIR_CONDITIONING:
-                await self._on_air_conditioning_event(event)
-            if event.topic == ServiceEventTopic.DEPARTURE:
-                await self._on_departure_event(event)
-            if event.topic == ServiceEventTopic.ODOMETER:
-                await self._on_odometer_event(event)
+            service_event_handlers: Dict[
```
dvx76, line 274: Would consider creating this in the `__init__` as `self.service_event_handlers`

custom_components/myskoda/coordinator.py
```diff
@@ -271,16 +271,19 @@ async def _on_mqtt_event(self, event: Event) -> None:
...
+                ServiceEventTopic.CHARGING: self._on_charging_event,
+                ServiceEventTopic.ACCESS: self._on_access_event,
+                ServiceEventTopic.AIR_CONDITIONING: self._on_air_conditioning_event,
+                ServiceEventTopic.DEPARTURE: self._on_departure_event,
+                ServiceEventTopic.ODOMETER: self._on_odometer_event,
+            }
+
+            handler = service_event_handlers.get(event.topic)
+            if handler:
+                await handler(cast(Any, event))
```
dvx76, line 286:
```suggestion
            if handler := service_event_handlers.get(event.topic):
                await handler(cast(Any, event))
```

Or, I think it might make sense to not be silent if somehow we get a eventopic which isn't in the mapping:

```suggestion
            try:
                await service_event_handlers[event.topic](cast(Any, event))
            except KeyError:
                _LOGGER.warning(f"Unhandled event topic {event.topic}")
```

Also don't love the use of `Any` and even less so having to cast to it 😅 . The lose useful type checking doing it this way. Previously if you'd accidentally try to call `self._on_charging_event(event)` with a access event instead of a charging event you'd get a pyright/pylance warning. I'm sure there is a better way to do this than what we currently have.

dvx76 (approved):

custom_components/myskoda/coordinator.py
WebSpider: Yeah, I understand what you mean, but I haven't found it yet :slightly_smiling_face:

WebSpider:
Spent some time today in trying to simplfy this, only to land up in complicated typing mayhem. Closing this and going to try again a bit later.

PR #727, opened 2025-03-26
Tell user S-PIN is missing when operating a lock

### LINKED ISSUE #578: Display a more informative message for entities that require an S-PIN
**Is your feature request related to a problem? Please describe.**
When entities require an S-PIN to operate and we do not have an S-PIN in the configuration, users do not understand what is going on

**Describe the solution you'd like**
Give these entities a state of "spin_required" (translatable) and make them unavailable so users know where to look

**Describe alternatives you've considered**
Keep things as is: New users will keep looking for support and wasting time to see what is going on
Not create the entities when there is no S-PIN: Even worse since users are even more lost

WebSpider:
Fixes #578

### DIFF
--- a/custom_components/myskoda/lock.py
+++ b/custom_components/myskoda/lock.py
@@ -8,6 +8,7 @@ from homeassistant.components.lock import (
     LockEntityDescription,
 )
 from homeassistant.core import HomeAssistant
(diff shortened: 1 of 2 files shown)


dvx76 (approved):
👍

PR #728, opened 2025-03-27
Add time left sensor for auxiliary heater

WebSpider:
Similar to the time left for AC, but this time for vehicles equipped with an auxiliary heater
TODO: Add translations

### DIFF
--- a/custom_components/myskoda/sensor.py
+++ b/custom_components/myskoda/sensor.py
@@ -67,6 +67,7 @@ async def async_setup_entry(
             SoftwareVersion,
             TargetBatteryPercentage,
             ClimatisationTimeLeft,
+            AuxHeaterTimeLeft,
         ],
         coordinators=hass.data[DOMAIN][config.entry_id][COORDINATORS],
         async_add_entities=async_add_entities,
@@ -731,7 +732,7 @@ class OutsideTemperature(MySkodaSensor):
 
 
 class ClimatisationTimeLeft(MySkodaSensor):
-    """Estimated time left until climatisation has reached its goal."""
+    """Estimated time left until climatisation via AC has reached its goal."""
 
     entity_description = SensorEntityDescription(
         key="estimated_time_left_to_reach_target_temperature",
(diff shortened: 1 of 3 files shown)


custom_components/myskoda/sensor.py
```diff
@@ -755,3 +756,30 @@ def native_value(self) -> int | None:  # noqa: D102
...
+        native_unit_of_measurement=UnitOfTime.SECONDS,
+        suggested_unit_of_measurement=UnitOfTime.MINUTES,
+        translation_key="aux_estimated_time_left_to_reach_target_temperature",
+    )
+
+    @property
+    def native_value(self) -> int | None:  # noqa: D102
+        if _aux := self.vehicle.auxiliary_heating:
+            if _aux.estimated_date_time_to_reach_target_temperature:
+                target_datetime = _aux.estimated_date_time_to_reach_target_temperature
```
dvx76, line 776: nit:

```suggestion
            if target_datetime := _aux.estimated_date_time_to_reach_target_temperature:
```

custom_components/myskoda/sensor.py
```diff
@@ -755,3 +756,30 @@ def native_value(self) -> int | None:  # noqa: D102
 
     def required_capabilities(self) -> list[CapabilityId]:
         return [CapabilityId.AIR_CONDITIONING]
+
+
+class AuxHeaterTimeLeft(MySkodaSensor):
+    """Estimated time left until climatisation via aux heater has reached its goal."""
+
+    entity_description = SensorEntityDescription(
+        key="aux_estimated_time_left_to_reach_target_temperature",
```
dvx76, line 765: Do we need entries in `icons.json` for this one and `estimated_time_left_to_reach_target_temperature` ?

dvx76 (changes requested):

custom_components/myskoda/sensor.py
WebSpider: Nice catch! I'll add them.

Issue #731, opened 2025-03-28
Wrong/old sensor info when Skoda server is unavailable


spockster1985:
### Please confirm:

- [x] I have searched the existing open and closed issues this problem
- [x] I have updated the integration to the latest available version

### Home Assistant installation and version

Core 2025.3.4

### Car model

Enyaq 85

### Latest working release

1.20.0

### Bug Description

When driving and now charging this morning, the battery percentage jumped back to the old value (64%) of the end of yesterdays driving. When it changed to 64%, I can see in the logs that the Skoda servers are unavailable (Internal server error 500)

![Image](https://github.com/user-attachments/assets/abcaceb9-e822-4152-86e4-655237e8e5ee)

### Debug Logs

```bash
2025-03-28 07:29:34.605 ERROR (MainThread) [myskoda.rest_api] Invalid status for GET request to /v1/charging/CAR_VIN: 500
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/myskoda/rest_api.py", line 101, in _make_request
    response.raise_for_status()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/client_reqrep.py", line 1161, in raise_for_status
    raise ClientResponseError(
    ...<5 lines>...
    )
aiohttp.client_exceptions.ClientResponseError: 500, message='Internal Server Error', url='https://mysmob.api.connect.skoda-auto.cz/api/v1/charging/CAR_VIN'
```

WebSpider:
Unfortunately, the MySkoda servers returning an error on a query that is otherwise valid, we cannot do much about. We are working on implementing a sensor that will indicate the time when the MySkoda servers last returned an error, to make you aware there is a problem. (#569)

Can you maybe tell us what you would like the value of this battery percentage sensor would like to be in this case, so we can see if we can adjust this in a way it will work for you?

spockster1985:
I understand that you can't fix the Skoda servers. 

But now the sensor shows an old wrong value. You can see from the screenshot that it manages to update the value in between the outages of the servers. But when the server is unavailable it suddenly shows a different value.

 If the sensor can just show that last good value.

WebSpider:
Yes, I understand.

The reason this is happening is because there are 2 sources for your battery information. One is coming in as a regular pull, the other one is coming from a push message.

[...]

PR #734, opened 2025-03-28
Add comparison for car_captured_timestamp

WebSpider:
As mentioned in #731 currently we always overwrite internal state with newly fetched information, even if this information is stale. There is no check.

This PR changes this by:

- Only handling events that have a timestamp that is more recent than the data we have from the car
- Only updating state when polling if the `car_captured_timestamp` is more recent than what we currently have in state

If the timestamp is missing, we assume we need to update.

### DIFF
--- a/custom_components/myskoda/coordinator.py
+++ b/custom_components/myskoda/coordinator.py
@@ -164,6 +164,13 @@ class MySkodaDataUpdateCoordinator(DataUpdateCoordinator[State]):
             self.vin, excluded_capabilities=excluded_capabilities
         )
 
+    def _is_more_recent(self, obj1, obj2) -> bool:
+        """Determines if an obj1 is newer than obj2. If timestamps are missing, obj1 wins."""
+        if not obj2.car_captured_timestamp or not obj1.car_captured_timestamp:
+            return True
+        else:
+            return obj2.car_captured_timestamp < obj1.car_captured_timestamp
+
     async def _async_update_data(self) -> State:
         vehicle = None
         user = None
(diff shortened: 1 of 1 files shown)


custom_components/myskoda/coordinator.py
```diff
@@ -164,6 +164,13 @@ async def _async_get_vehicle_data(self) -> Vehicle:
             self.vin, excluded_capabilities=excluded_capabilities
         )
 
+    def _is_more_recent(self, obj1, obj2) -> bool:
```
dvx76, line 167: Fun fact: `Health` has `captured_at` instead of `car_captured_timestamp` and `Maintenance` doesn't have either (but `Maintenance.MaintenanceReport` also has `captured_at`) 😭 

I would also much prefer to implement this at the class level, using `__gt__` (or if we think that's abusive, something like a `is_newer()` method). So that you can just do `if health_info > vehicle.health` (or `if health.is_newer(vehicle.health)`)

dvx76 (changes requested):
`MySkodaDataUpdateCoordinator._async_update_data()` is still going to end up restoring the 'old' value ☹️  (see https://github.com/skodaconnect/homeassistant-myskoda/issues/731#issuecomment-2761795008)

WebSpider:
> `MySkodaDataUpdateCoordinator._async_update_data()` is still going to end up restoring the 'old' value ☹️ (see [#731 (comment)](https://github.com/skodaconnect/homeassistant-myskoda/issues/731#issuecomment-2761795008))

I'll make the changes there as well!

custom_components/myskoda/coordinator.py
WebSpider: Nice suggestion!

I'll work on implementing this probably after the weekend.

dvx76:
Potential alternative solution, based on https://github.com/skodaconnect/homeassistant-myskoda/issues/731#issuecomment-2761795008, in https://github.com/skodaconnect/myskoda/pull/407

Let me know what you think.

WebSpider:
> Potential alternative solution, based on [#731 (comment)](https://github.com/skodaconnect/homeassistant-myskoda/issues/731#issuecomment-2761795008), in [skodaconnect/myskoda#407](https://github.com/skodaconnect/myskoda/pull/407)
> 
> Let me know what you think.

Reviewed. It maybe doesn't cover everything this PR does, but I'll rework this as/if needed after it merges.

Issue #411, opened 2024-12-05
Report fixtures via HA diagnostics


dvx76:
See if we can use https://www.home-assistant.io/integrations/diagnostics/ to make it easier for users to report fixtures

WebSpider:
This looks very doable, looks like all we need to do is add generating fixtures to an override of `async_get_device_diagnostics`

prvakt:
Made some tests on my place and seems its working.. Need to play with text formatting a little but I can implement it.
Just a question - in case of multiple vehicles - there is separate coordinator for each one?

prvakt:
#415

PR #746, opened 2025-04-08
Refactor MySkoda and MQTT update logic

dvx76:
This moves all update logic out of the coordinator and into the MySkoda lib.

The coordinator can still subscribe to MQTT events but they are not used for anything and just get exposed in corresponding entities.

The coordinator can now subscribe to 'update' events which MySkoda will callback when any Vehicle data was updated for whatever reason (including when it, the MySkoda lib, got an MQTT event). When receiving an update callback the coordinator just has to call async_set_updated_data and pass in the current Vehicle data.

Goes with https://github.com/skodaconnect/myskoda/pull/411

Draft, still testing and fine-tuning this.


dvx76:
This is now ready 🚀

WebSpider (approved):
Lgtm! Thanks for all of this work! 🎉

Issue #741, opened 2025-04-04
Start loading not possbible


gerihouse:
### Please confirm:

- [x] I have searched the existing open and closed issues this problem
- [x] I have updated the integration to the latest available version

### Home Assistant installation and version

2025.3.4

### Car model

Enyaq 80

### Integration version

1.22.0

### Bug Description

Using the switch to start the loading is not working. After some time switch is back to off state. 

In the Skoda App it displays an error after some time that "loading is not possible at the moment"

If I start the loading in the Charger App it works. Afterwards turning off the loading in HA works.

### Debug Logs

```bash

```

dvx76:
This might be coincidence if it can't be reproduced multiple times. The API is not 100% reliable. Either way, without logs there is nothing we can do to investigate this.

gerihouse:
OK thanks for your fast answer. Next time I will activate logging.

gerihouse:
@dvx76 now I have a debug log:

2025-04-11 13:26:28.255 DEBUG (Dummy-1) [myskoda.rest_api] Starting charging for vehicle TMBJC7NY0PF077641
2025-04-11 13:26:28.769 DEBUG (Dummy-1) [myskoda.mqtt] Waiting for operation start-charging complete.
2025-04-11 13:26:28.779 DEBUG (Dummy-1) [myskoda.mqtt] Received PUBLISH (d0, q0, r0, m0), '0487bfe5-2e30-43ec-860a-2b2b52cd4784/TMBJC7NY0PF077641/operation-request/charging/start-stop-charging', ...  (161 bytes)
2025-04-11 13:26:28.781 DEBUG (Dummy-1) [myskoda.mqtt] Message (operation-request) received for TMBJC7NY0PF077641 on topic charging/start-stop-charging: b'{"version":1,"operation":"start-charging","status":"IN_PROGRESS","traceId":"caa437782c03a7947860ce210653cb6a","requestId":"b2de58a1-13fd-4f17-90cf-924459934e6f"}'
2025-04-11 13:26:28.781 DEBUG (Dummy-1) [myskoda.mqtt] An operation 'start-charging' is now in progress. Trace id: caa437782c03a7947860ce210653cb6a
2025-04-11 13:26:28.782 DEBUG (Dummy-1) [custom_components.myskoda.coordinator] Manually updated myskoda data
[...]

PR #750, opened 2025-04-14
Fix and refactor switch onoff logging

dvx76:
Related to #741

When turning a switch on or off would fail an errors with the details
was logged, followed by a success message.

Also includes minor refactor to de-duplicate some of the code, as well as
align these functions and their logging across all switches.

### DIFF
--- a/custom_components/myskoda/switch.py
+++ b/custom_components/myskoda/switch.py
@@ -2,7 +2,7 @@
 
 import logging
 from datetime import timedelta
-from typing import Any
+from typing import Any, Coroutine
 
 from homeassistant.components.switch import (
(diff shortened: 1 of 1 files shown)


WebSpider (commented):
Would it make sense to put this logic into all entities that can flip, so not only switches, but also buttons?

dvx76:
> Would it make sense to put this logic into all entities that can flip, so not only switches, but also buttons?

Probably, yeah. I only looked at switch :)

Happy to take a look at it but won't have time this weekend so maybe open an issue and I'll do it later and in the meantime this can be merged :)

WebSpider (approved):
Lgtm

PR #753, opened 2025-04-18
Bump version to 1.23.0

dvx76:
Drop version from pyproject.toml since it's not being used there (we don't package anything)

Planning to do a pre-release of this one first to confirm behavior after the large refactor.


WebSpider (approved):
Lgtm

PR #756, opened 2025-04-20
Improve terms-and-conditions login error

### LINKED ISSUE #676: Add details when logging TermsAndConditionsError, MarketingConsentError
From #674, we should include the exception text when logging `TermsAndConditionsError` and `MarketingConsentError` (which do include the value of the location header), as well the the URL of the Skoda portal (what we also report in the issue which gets created).

In `async_setup_entry()`

```python
    except (TermsAndConditionsError, MarketingConsentError) as exc:
        _LOGGER.error(
            "Change to terms and conditions or consents detected while logging in. Please log into the MySkoda app (may require a logout first) to access the new Terms and Conditions. This HomeAssistant integration currently can not continue."
        )
        async_create_tnc_issue(hass, entry.entry_id)
```

dvx76:
Fixes #676


WebSpider (approved):
Lgtm

Issue #404, opened 2024-12-04
Outside temperature sensor showing incorrect value


Tomanic-glitch:
It seems that the new Outside Temperature sensor in v1.14.1 is giving wrong value compared to the old Skoda Connect integration for my Skoda Kodiaq II.

Myskoda:
![Myskoda](https://github.com/user-attachments/assets/94ce592e-0340-41f8-91ea-a0a1e75c9723)

Skoda Connect:
![Skoda_connect](https://github.com/user-attachments/assets/fdc08a27-0089-4d36-992a-556c5e69f380)

WebSpider:
> It seems that the new Outside Temperature sensor in v1.14.1 is giving wrong value compared to the old Skoda Connect integration for my Skoda Kodiaq II.
> 
[...]

PR #764, opened 2025-04-29
Switch from poetry to uv

dvx76:
Based on #763 so I don't revert the myskoda version.

Follows the same changes applied to the myskoda project:
* https://github.com/skodaconnect/myskoda/pull/394
* https://github.com/skodaconnect/myskoda/pull/404/files

Usage is very similar to poetry:
* `poetry lock` -> `uv lock`
* `poetry install --all-extras` -> `uv sync --all-extras`
* `poetry run` -> `uv run`

```
❯ uv sync --all-extras
Resolved 128 packages in 0.50ms
Audited 117 packages in 0.05ms
```

```
❯ uv run pre-commit run --hook-stage manual --all-files
ruff.....................................................................Passed
ruff-format..............................................................Passed
pyright..................................................................Passed
uv-lock..................................................................Passed
```


.github/workflows/lint.yaml
```diff
@@ -11,22 +11,21 @@ jobs:
...
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: "0.6.17"
 
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
           python-version-file: pyproject.toml
-          cache: "poetry"
```
WebSpider, line 24: Would it make sense to keep caching the Python setup? Speed up CI

pyproject.toml
```diff
@@ -1,27 +1,20 @@
-[tool.poetry]
+[project]
 name = "homeassistant-myskoda"
+version = "0.0.1"
```
WebSpider, line 3: Is this 0.0.1 in `myskoda` as well? I thought we had set this to 0.0.0 there.

pyproject.toml
```diff
@@ -1,27 +1,20 @@
...
-[build-system]
-requires = ["poetry-core"]
-build-backend = "poetry.core.masonry.api"
+[dependency-groups]
+dev = ["homeassistant-stubs>=2025.3.2", "ruff ~=0.11.7", "pyright ~=1.1.400"]
 
 [tool.pyright]
 reportIncompatibleVariableOverride = "none"
+venvPath = "."
+venv = ".venv"
```
WebSpider, line 20: Probably needs an update of `.gitignore` as well

poetry.lock
WebSpider, line 1: This needs removed i think

WebSpider (commented):
Yeah lets use the same setup for both. Some minor remarks on this PR, but feel free to proceed.

dvx76:
> Yeah lets use the same setup for both. Some minor remarks on this PR, but feel free to proceed.

Yeah I wasn't sure what to use, because there's no `cache: "uv"` (see https://github.com/actions/setup-python#caching-packages-dependencies). But I think maybe `cache: "pip"` might work.

Alternatively there is a uv-specific action from astral: https://github.com/astral-sh/setup-uv#do-i-still-need-actionssetup-python-alongside-setup-uv . I'll try both.

.github/workflows/lint.yaml
dvx76: Yeah I wasn't sure what to use, because there's no cache: "uv" (see [actions/setup-python#caching-packages-dependencies](https://github.com/actions/setup-python#caching-packages-dependencies)). But I think maybe cache: "pip" might work.

Alternatively there is a uv-specific action from astral: [astral-sh/setup-uv#do-i-still-need-actionssetup-python-alongside-setup-uv](https://github.com/astral-sh/setup-uv#do-i-still-need-actionssetup-python-alongside-setup-uv) . I'll try both.

poetry.lock
dvx76: Yeah I must have messed up my rebase.

pyproject.toml
dvx76: Updated

PR #767, opened 2025-04-30
feat: work around API sending incorrect mileage

### LINKED ISSUE #766: Dont let Odometer value decrease during cloud crash
### Please confirm:

docker 2025.3.2

### Integration version

1.23.1

### Car model

Citigo IV

### Feature Description

Hi. as mentioned at Discord, I have seen a phenomenon with odometer decrease when there is probably some crash of the cloud servers like today. The sensor reports lower value, can this be checked first and filtered? I have a Meter for daily distance and it reports values that are to high..
see the Odometer from today:

Image

thanks for your consideration.
Maybe I can gate the decrease of my custom Daily Distance Meter myself somehow?

### MySkoda App Behavior

App was without response for a few hours today.

dvx76:
Fixes #766 (maybe)

The Skoda API sometimes returns an incorrect (old) mileage. This workaround inspects the last recorded value, if there is one, and uses that if it happens to be larged than the value returned by the API.

To test I added a logline for `last_value` and it seems to get the expected value

```
2025-04-30T19:04:32.111307325Z 2025-04-30 19:04:32.102 INFO (MainThread) [custom_components.myskoda.sensor] XXXX LAST_VALUE = 54606
```

### DIFF
--- a/custom_components/myskoda/sensor.py
+++ b/custom_components/myskoda/sensor.py
@@ -534,11 +534,26 @@ class Mileage(MySkodaSensor):
 
     @property
     def native_value(self) -> int | None:  # noqa: D102
+        """Calculate the mileage_in_km.
+
+        The API sometimes erroneously returns an old, lower value. Mileage should never go down.
+        To work around this we inspect the last state, if there is one, and return that if it is
+        larger than the value from the API.
+        """
+        last_value = 0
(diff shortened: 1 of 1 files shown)


WebSpider (approved):
Lgtm.

PR #769, opened 2025-05-01
fix: prevent value/type error when setting last mileage value

dvx76:
Introduced in #767

Woops.

Can happen on startup.

```
2025-05-01T09:36:22.033377902Z   File "/config/custom_components/myskoda/sensor.py", line 548, in native_value
2025-05-01T09:36:22.033395550Z     last_value = int(last_state.state)
2025-05-01T09:36:22.033412235Z ValueError: invalid literal for int() with base 10: 'unavailable'
```


WebSpider (approved):
Lgtm.

Is this the only variable that can be unavailable/None on init?

Issue #762, opened 2025-04-27
Wake car


TimonTrukker:
### Please confirm:

- [x] I have searched the existing open and closed issues this problem
- [x] I have updated the integration to the latest available version

### Home Assistant installation and version

2025.4.4

### Integration version

1.23.0

### Car model

Kodiaq

### Feature Description

I miss the possibility to wake the car to refresh the state of it. My car only has a combustion engine and I think for that reason it keeps communication at a minimum. The result is that if I do not lock the door immediately after turning of the car, the state in HA (and the my skoda app) stays unlocked until I force a refresh from the my skoda app. With the old add-on, I solved this with an automation which forced refreshing the state by the car whenever the doors are unlocked for a while to make sure this is correct. With this add-on, this feature is not available anymore and I would love to have it back. 

### MySkoda App Behavior

In the skoda app, simply opening the app causes the car to wake (within some constraints), but for the add-in it would be nice is this is button (or maybe a switch)

WebSpider:
Duplicate of #687

WebSpider:
I've been on holiday so a reply took a while.

The trouble here is that for some models, the car is really restrictive in the amount of times you can wake the car without requiring physical presence to turn on the ignition/contact. All communication from the car to myskoda stops as far as we can detect. We've seen this happen in the old `skodaconnect` integration. 

In order to prevent ever ending up in this state (which can happen easily if you for instance have a Citigoe, it allows 3 wakeups), we never implemented this.

I guess we could make this a gated feature, meaning you have to enable the it in the configuration (where you also enter the S-PIN), and then a (diagnostic/hidden) button gets created that does a wakeup when the car is not already online or waking up.

Does that help your case?

@dvx76 anything to add?

TimonTrukker:
Hi,

That would help my case.

In my case, I rather have an update with a correct state then an incorrect state.

Atm, I am waking the car by opening the myskoda app manually, so for me, the only change is that I can automate the wake 😎.

In any case, thanks for all the effort, both historical an future 👍

dvx76:
An extra checkbox in the config sounds good to me. Alternatively, only making it available via a service call (now called 'action' in HA), if that's any easier.

TimonTrukker:
I agree, a service call (action) would be sufficient.

PR #777, opened 2025-05-06
feat: add button to wake up the car

dvx76:
Fixes: #762

Disabled by default to limit accidental use.

Adding service calls (actions) looked like it would actually be more work/complexity so I ended up going for a third option after the discussion in #762: directly add a new button, but keep it disabled by default. This means the use has to first intentionally find, update and enable the entity before being able to use it.

Seems to work ok on my Enyaq although I can't tell if it really does wake the car

```
2025-05-06T07:20:48.734252002Z 2025-05-06 07:20:48.729 DEBUG (MainThread) [myskoda.myskoda] Trace: POST https://mysmob.api.connect.skoda-auto.cz/api/v1/vehicle-wake - response: 202 (0 bytes)
```

I didn't see any MQTT event but I'm not seeing _any_ events right now so I think the MQTT broker might be down...

### DIFF
--- a/custom_components/myskoda/translations/en.json
+++ b/custom_components/myskoda/translations/en.json
@@ -354,6 +354,9 @@
             },
             "flash": {
                 "name": "Flash"
+            },
+            "wakeup": {
+                "name": "Wake Up Car"
             }
         }
     },
--- a/custom_components/myskoda/button.py
+++ b/custom_components/myskoda/button.py
@@ -32,7 +32,7 @@ async def async_setup_entry(
 ) -> None:
     """Set up the button platform."""
     add_supported_entities(
-        available_entities=[HonkFlash, Flash],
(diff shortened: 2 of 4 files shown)


custom_components/myskoda/translations/en.json
```diff
@@ -354,6 +354,9 @@
             },
             "flash": {
                 "name": "Flash"
+            },
+            "wakeup": {
+                "name": "Flash"
```
WebSpider, line 359: Copy/paste mistake? 😉

WebSpider (changes requested):
Translation seems a bit off. Otherwise lgtm!

custom_components/myskoda/translations/en.json
dvx76: Didn't save the file in my VSCode 🤦

WebSpider (approved):
Lgtm

PR #798, opened 2025-05-13
feat: add 3 new binary sensors using connection-status API endpoint

### LINKED ISSUE #751: Car/communication status
### Please confirm:

2025.4.2

### Integration version

1.22.0

### Car model

Enyaq iV 80

### Feature Description

It would be nice to have the communication status available in order to find out for instance if the communication is cut and no events are sent anymore and the care won't accept any commands anymore from HA/Myskoda. This happens everytime the 12V battery is in a low charge state.
It goes into some sort of saving mode and communication is cut.

### MySkoda App Behavior

In the Myskoda App it shows the status in the top left corner as a colored dot. if you press the information sign you get the legend of all the colors.

Image
Image

WebSpider:
This adds 3 new `BinarySensor`s, representing the data that we receive from the new endpoint.

`vehicle_reachable`: On if the vehicle can be reached by the MySkoda servers
`vehicle_battery_protection`: On if the vehicle currently is in Battery Protection mode
`vehicle_in_motion`: On if the vehicle currently is in motion. This seems to be a duplication of the motion detection we do in the `device_tracker` currently, we'll have to see how to move forward with this as/when we have more experience on the matter.

Since we do not have much information about the timeliness, availability and correctness of the endpoint, the default translation now shows this as "beta".

Fixes #751

### DIFF
--- a/custom_components/myskoda/binary_sensor.py
+++ b/custom_components/myskoda/binary_sensor.py
@@ -5,6 +5,7 @@ from homeassistant.components.binary_sensor import (
     BinarySensorEntity,
     BinarySensorEntityDescription,
 )
+from homeassistant.const import EntityCategory
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
 from homeassistant.helpers.typing import DiscoveryInfoType  # pyright: ignore [reportAttributeAccessIssue]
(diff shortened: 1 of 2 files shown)


dvx76 (approved):
Great work! Let's see how the data from this compared with the device tracker.

PR #800, opened 2025-05-15
chore(i10n): remove binary sensor state translations from localization

### LINKED ISSUE #384: remove sensor state values from translation

WebSpider:
This removes the state translations of binary sensors, since they are equal to the default translations inside HA.
Once the strings are removed from the reference language (English), our translation tool should provide us with an automatic PR to clean it up in all languages.

Thanks @ChristophCaina for pointing this out and providing the first fix, however as indicated in #576 he currently has other things to do.

Fixes #384

### DIFF
--- a/custom_components/myskoda/translations/en.json
+++ b/custom_components/myskoda/translations/en.json
@@ -33,102 +33,46 @@
     "entity": {
         "binary_sensor": {
             "bonnet_open": {
-                "name": "Bonnet",
-                "state": {
-                    "off": "Closed",
-                    "on": "Open"
-                }
+                "name": "Bonnet"
             },
             "charger_connected": {
-                "name": "Charger Connected",
-                "state": {
-                    "off": "Unplugged",
-                    "on": "Plugged in"
-                }
+                "name": "Charger Connected"
             },
             "charger_locked": {
-                "name": "Charge Lock",
-                "state": {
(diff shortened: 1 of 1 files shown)


ChristophCaina:
Hi, sorry for not comming Back to this...
There was a lot of stress with several Things privately and at Work - and tbh I wasn't in the mood of working in this.

I wanted to Look Into this in the upcomming week when I'm on Holidays

dvx76:
> Hi, sorry for not comming Back to this... There was a lot of stress with several Things privately and at Work - and tbh I wasn't in the mood of working in this.
> 
> I wanted to Look Into this in the upcomming week when I'm on Holidays

Hey @ChristophCaina , you probably meant to comment on #576 I guess 😅 

Don't worry about it. This is a side-project for all of us and everyone has different priorities at different times. Work on this when you feel like it 😉

PR #807, opened 2025-05-21
chore: instruct hacs to download the zip not checkout the repo

WebSpider:
Downloading the zip is faster than checkout of the repo, and we create it in CI a while now.


dvx76 (approved):
🚀

Issue #140, opened 2024-10-27
myskoda.mqtt.ConcurrentSubscribeError causes HA Core to get unresponsive


ChristianForstner:
After updating from version 1.5.0 to 1.6.2, this error occurs:

```
2024-10-26 14:19:50.434 ERROR (MainThread) [myskoda.mqtt] Failed to connect to MQTT.
2024-10-26 14:19:55.537 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved (None)
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/myskoda/mqtt.py", line 150, in connect
    await self._subscribe_to_topic(f"{user_id}/{vin}/operation-request/{topic}")
  File "/usr/local/lib/python3.12/site-packages/myskoda/mqtt.py", line 260, in _subscribe_to_topic
    raise ConcurrentSubscribeError
myskoda.mqtt.ConcurrentSubscribeError
```

This `myskoda.mqtt.ConcurrentSubscribeError` causes HA core to get unresponsive.

```
2024-10-26 14:29:18.641 ERROR (MainThread) [homeassistant.components.hassio.handler] Client error on /core/info request Cannot connect to host 172.30.32.2:80 ssl:default [No such file or directory]
2024-10-26 14:29:18.648 ERROR (MainThread) [homeassistant.components.hassio.coordinator] Error fetching hassio data: Error on Supervisor API: 
```

```
2024-10-26 15:14:20.250 ERROR (MainThread) [homeassistant.components.hassio.handler] Client error on /os/info request Cannot connect to host 172.30.32.2:80 ssl:default [No such file or directory]
2024-10-26 15:14:20.252 WARNING (MainThread) [homeassistant.components.hassio] Can't read Supervisor data: 
```

```
2024-10-26 15:19:18.639 ERROR (MainThread) [homeassistant.components.hassio.handler] Client error on /info request Cannot connect to host 172.30.32.2:80 ssl:default [No such file or directory]
2024-10-26 15:19:18.647 ERROR (MainThread) [homeassistant.components.hassio.coordinator] Error fetching hassio data: Error on Supervisor API: 
```

```
2024-10-26 15:20:28.354 ERROR (MainThread) [homeassistant.components.homeassistant_alerts.coordinator] Timeout fetching homeassistant_alerts data
```

At some point, also a `DNSError` occurs:
```
[...]

PR #833, opened 2025-06-11
chore: remove deprecated MySkoda ssl_context

dvx76:
This doesn't do anything anymore since asyncio-paho was replaced with aiomqtt in skodaconnect/myskoda#140

(import order changes due to setting up VSCode to let ruff auto-organize imports on save)

### DIFF
--- a/custom_components/myskoda/__init__.py
+++ b/custom_components/myskoda/__init__.py
@@ -5,32 +5,30 @@ from __future__ import annotations
 import logging
 
 from aiohttp import ClientResponseError, InvalidUrlClientError
-
 from homeassistant.const import Platform
 from homeassistant.core import HomeAssistant
 from homeassistant.exceptions import ConfigEntryAuthFailed, ConfigEntryNotReady
 from homeassistant.helpers import entity_registry as er
 from homeassistant.helpers.aiohttp_client import async_create_clientsession
-from homeassistant.util.ssl import get_default_context
+
 from myskoda import (
-    MySkoda,
     AuthorizationFailedError,
+    MySkoda,
 )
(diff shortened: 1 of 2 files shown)


WebSpider:
I remember we added this because if we didnt, HA would complain about blocking calls when reading the CA list from disk.

I wonder if that behaviour has changed

dvx76:
> I remember we added this because if we didnt, HA would complain about blocking calls when reading the CA list from disk.
> 
> I wonder if that behaviour has changed

I also remember something to that effect but I tracked the addition of it in the MQTT Client code to https://github.com/skodaconnect/myskoda/commit/28ebb06609f04065c92b550208c11d441f7d4e12#diff-29d6a3414735d738e7217d1c05e1f0b8d1d0447f41fd1d68e2e704e770649b00 but that didn't make me any wiser 😅 

It's definitely working right now with the context we create locally: https://github.com/skodaconnect/myskoda/blob/71f67b48d38adbf3b43c0fda1695ac0c53c030f6/myskoda/mqtt.py#L38 - I suppose it works because it's done at import time instead of at runtime.

I can try replacing that with the HA context we're being passed in.

WebSpider (changes requested):
Small optimization, otherwise lgtm.

dvx76:
> I can try replacing that with the HA context we're being passed in.

Works fine with the HA context (`get_default_context()`). On the one hand if we don't _need_ it (and obviously we don't since it's working right now for everyone) I'd rather just have less code. But using the context from HA _seems like the right thing to do_ ... it's also what all other core integrations seem to do.

So I'll close this and instead switch back to using the HA context.

PR #851, opened 2025-07-07
feat: implement parking position device tracker (beta)

### LINKED ISSUE #824: MySkoda API v3 parking GPS address
### Please confirm:

2025.4.1

### Integration version

1.22.0

### Car model

Skoda Enyaq 85 - 2026 

### Feature Description

I want access to GPS coordinates in the HA integration.
I set up an old android phone with linage OS and intercepted the traffic from the MySkoda app.

The endpoint to get the coordinates for the new version of the API is as follows:
`https://mysmob.api.connect.skoda-auto.cz/api/v3/maps/positions/vehicles/<VIN>/parking` 

This endpoint uses the same bearer authentication used by v1.
The returned data is as follows:
```
{
  "parkingPosition": {
    "gpsCoordinates": {
      "latitude": xx.xxxxx,
      "longitude": yy.yyyyy
    },
    "formattedAddress": "bla bla bla"
  }
}
```

### MySkoda App Behavior

_No response_

WebSpider:
This implements the output of the new parking position data as a device tracker.

We currently do not have enough data available to know if this sensor is available for all vehicles that have `PARKING_POSITION` capability available (that's 100% of our fixtures right now), so this will be implemented in beta releases first.

Closes #824

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1102,7 +1102,7 @@ wheels = [
 
 [[package]]
 name = "myskoda"
-version = "2.3.0"
+version = "2.3.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "aiohttp" },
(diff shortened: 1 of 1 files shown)


custom_components/myskoda/device_tracker.py
```diff
@@ -119,3 +127,65 @@ def location_name(self) -> str | None:
...
         return [CapabilityId.PARKING_POSITION]
+
+
+class ParkingPositionTracker(MySkodaEntity, TrackerEntity):
+    """Tracker for the last known parking position."""
+
+    def __init__(self, coordinator: MySkodaDataUpdateCoordinator, vin: str) -> None:
+        self.entity_description = TrackerEntityDescription(
+            name="Parking",
+            key=f"{vin}_parking_position",
```
dvx76, line 138: Was there a reason we left the `vin` in the key of the `DeviceTracker`? None of the other entities have it.

(Not asking you to change it in this PR - just wondering about it)

dvx76 (approved):
LGTM. Doesn't need to be in this PR since we're marking this as beta, but eventually we should add this in https://github.com/skodaconnect/homeassistant-myskoda/blob/main/docs/entities.md#device-tracker and explain the difference with the existing device tracker (which I see is incorrectly documented there 🙈 )

custom_components/myskoda/device_tracker.py
WebSpider: Good spot!

I'll check this before committing. We only need the vin for the unique_id, so i think we can drop it here.

Renko72:
FYI to the following sentence "we currently do not have enough data available to know if this sensor is available for all vehicles" 
i have installled the beta from MySkoda, but with my CAR Skoda ELROQ (Year 2025) with SW Version 5.4 i get no Parking_Position DATA

PR #856, opened 2025-07-28
fix: rename unique_id of binary sensor locks from 'locked' to 'lock'

### LINKED ISSUE #853: Lock Sensor State is Inverted
### Please confirm:

HAOS 2025.7.1

### Car model

Elroq 50

### Integration version

1.28.3

### Latest working release

_No response_

### The Problem
The `binary_sensor` for the vehicle lock appears to have inverted logic, which makes creating automations confusing. The entity is named `binary_sensor.skoda_elroq_vehicle_locked`, but its state is `on` when the vehicle is unlocked.

### Actual vs. Expected Behavior

* **Actual Behavior:**
    * When the car is physically **LOCKED**, the sensor state is `off`.
    * When the car is physically **UNLOCKED**, the sensor state is `on`.

* **Expected Behavior:**
    * Given the entity name `...vehicle_locked`, the expected behavior is for the state to be `on` when the vehicle is locked.

### Debug Logs

```bash

```

WebSpider:
This PR changes the suffix in the unique_id of all LOCK type binary sensors from "locked" to "lock" to prevent confusion.

This still needs some testing to see if we migrate correctly.

Fixes #853

### DIFF
--- a/custom_components/myskoda/__init__.py
+++ b/custom_components/myskoda/__init__.py
@@ -287,7 +287,12 @@ async def async_migrate_entry(hass: HomeAssistant, entry: MySkodaConfigEntry) ->
 
             for entity in entry_entities:
                 if entity.unique_id in old_entities:
-                    new_unique_id = entity.unique_id.replace("locked", "lock")
+                    if entity.unique_id.endswith(("charger_locked", "doors_locked")):
+                        new_unique_id = entity.unique_id.replace("locked", "lock")
+                    else:
+                        new_unique_id = entity.unique_id.replace(
+                            "locked", "vehicle_lock"
+                        )
                     _LOGGER.debug(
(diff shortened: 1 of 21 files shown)


dvx76 (approved):
LGTM, pending test result

WebSpider:
I am going to remove the "breaking" label here, since after testing it has become clear that:

- No new entities are made
- No existing entities are removed or abandoned, so device selections from the UI/yaml still work
- No `entity_id`'s are changed, so entity selections (often used in blueprints/manually written automations) still work

What is left for the end user is:

- If they select anything based on the `unique_id`: This has changed. We should put this in the changelog
- If they want to rename the `entity_id` to now be called `lock` instead of `locked`, they can do so themselves, taking into account any automations that reference it. With the `entity_id` being translated by default, the migration code would increase a lot in complexity.

PR #915, opened 2025-10-01
Show correct value when battery% is 0

### LINKED ISSUE #913: Battery level 0% is reported as unavailable
### Please confirm:

HAOS 2025.9.4

### Car model

Skoda Superb iV 2025

### Integration version

v1.28.8

### Latest working release

_No response_

### Bug Description

In hybrid cars, the battery level 0% is valid level as we can run out of battery and continue driving on gasoline. When level is 0, the battery level sensor reports that the value is "Unavailable". This makes some computation script/algorithms to misbehave (e.g. the one that schedules charges on low electricity prices).
Do you think, there can be distinction between the sensor being 0% or truly unavailable?

### Debug Logs

```bash

```

WebSpider:
Fixes #913

### DIFF
--- a/custom_components/myskoda/sensor.py
+++ b/custom_components/myskoda/sensor.py
@@ -215,7 +215,7 @@ class BatteryPercentage(ChargingSensor):
     @property
     def native_value(self) -> int | None:  # noqa: D102
         if status := self._status():
-            if status.battery.state_of_charge_in_percent:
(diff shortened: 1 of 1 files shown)


fursov:
Should also this line be changed same way?
https://github.com/skodaconnect/homeassistant-myskoda/blob/5b4b1c575a32144b74e6181004cbabd68512caec/custom_components/myskoda/device_tracker.py#L187

PR #939, opened 2025-11-09
chore(migration): Add migration downgrade notice. 

WebSpider:
Any schema downgrade inside schema v2 is not supported


WebSpider:
Superseded by #940

PR #940, opened 2025-11-10
feat(auth): Store and reuse oidc2 refresh_token

WebSpider:
This enhances the integration by:

- Allowing storage of the current `refresh_token`
- Using the `refresh_token` as primary authentication method instead of username/password
- Updating the `refresh_token` as needed when shutting down

Also:

- The migration code for schema versions has been changed so that multiple migrations can take place in 1 run
- We no longer fail when a user in on a higher minor schema than we support (this enhances the `beta` experience)


WebSpider:
Before release I'll make it use a proper version of the `myskoda` library :blush:

WebSpider:
Tested this on my dev environment, 2025.10.1

PR #962, opened 2025-12-10
Fix: Readonly mode did not block locking/unlocking or heater actions

### LINKED ISSUE #944: Button FLASH & HONK_FLASH
### Please confirm:

2025.11.1

### Car model

SKODA KODIAQ

### Integration version

1.28.9

### Latest working release

1.28.8

### Bug Description

The FLASK and HONK_FLASH buttons are not present. They are present and functional from the app.

### Debug Logs

```bash

```

racoon2code:
While working on Issue #944, I discovered that the integration did not fully respect readonly mode.
Despite readonly, it was still possible to:

- lock the vehicle

- unlock the vehicle

- enable the auxiliary heater

Readonly mode should prevent any vehicle-controlling actions, so this behavior was incorrect.

Fix

This PR implements proper readonly enforcement:

- all write operations now check the CONF_READONLY flag before executing

- service handlers block commands while readonly mode is active

- entity availability correctly reflects readonly status

- the readonly option is normalized to a boolean to avoid incorrect evaluations

Readonly mode now consistently prevents any action that modifies the vehicle state.

Why this was a bug

Readonly mode implies strict read-only behavior, but the integration still allowed safety-relevant commands.
This was caused by missing checks and unreliable readonly evaluation.

Testing

- readonly enabled → all write actions correctly blocked

- readonly disabled → normal behavior restored

- entities update availability as expected

- all pre-commit checks (Ruff, format, Pyright) pass

References
Fixes #944


WebSpider:
Reading the remark you placed in #944 i understand why you did it this way, and this would certainly be consistent.. However, I think we can improve the current implementation while we are at it. Unavailable or abandoned entitites just create a mess inside HA

racoon2code:
@WebSpider Thanks for the feedback!

I tried to adjust the implementation as you suggested so that the entities remain available in readonly mode and only the actions are blocked. However, this leads to the following UX issue, using the lock/unlock entity as an example:

- The entity is present and the lock/unlock control is still shown.

- When you press the button in readonly mode, nothing happens because the command is blocked internally.

- If readonly was enabled by accident, it’s quite hard for the user to understand why the action doesn’t work – the only hint is in the logs.

From a UX perspective, I’m not sure this is actually better than hiding/un-creating the entity in readonly mode. When I created this PR, I tried to align the behavior with other entities in the integration such as Honk & Flash and the EV climate controls, where the entities simply aren’t exposed in readonly mode.

If we want to keep the entities visible but disable the actions when readonly is enabled, I’m happy to implement that. In that case I would suggest we do it consistently for (almost) all vehicle-controlling entities, not just the two I touched in this PR, so the behavior is predictable across the board.

Let me know which direction you prefer:

A) Entities are hidden / not created in readonly mode (current behavior in some places like Honk & Flash, EV climate), or

B) Entities stay visible, but their actions are disabled in readonly mode – implemented consistently across all write-capable entities.

I can then adjust the PR accordingly. 🙌

racoon2code:
@WebSpider  
I've updated the behavior as requested.
All relevant entities now stay active and available in readonly mode, but any write action (locking, unlocking, heater, climate, etc.) is cleanly blocked and raises the readonly exception. This keeps everything visible while still enforcing the intended restrictions.

One additional thought: the exception currently uses a fixed text string. It might be nicer to replace that with a proper Home Assistant translation key so the message can be localized. That would require updating all translation files, so it might make sense to track that separately in a new issue.

WebSpider:
> One additional thought: the exception currently uses a fixed text string. It might be nicer to replace that with a proper Home Assistant translation key so the message can be localized. That would require updating all translation files, so it might make sense to track that separately in a new issue.

Feel free to add that to this PR. You only need to set up the translation in the proper section of `en.json` and when merged, it will become translatable for all languages we currently support.

racoon2code:
@WebSpider 
Thanks! I’ve implemented this now.
The ServiceValidationError was updated to use a translation_key, and the corresponding entry was added to the exceptions section in en.json.

WebSpider:
Thanks for this!

Sorry about the late response, life has been kind of  busy.

All in all, this looks like a good PR. I will do some testing with it after release of 1.29.0, to see if the UX makes sense now.

racoon2code:
That makes sense to me. Reverting the UI state after a blocked action sounds like a good UX improvement. I’ll look into how we can do this cleanly and get back to you.

WebSpider:
If this takes a bit longer, lets get this merged in and do the UI improvements later @racoon2code

racoon2code:
Hi @WebSpider,

sorry for the delay. At the moment I can’t really estimate how long the UI improvement will take, as I haven’t had the chance to look into it yet due to the holidays and spending time with family.

I’m perfectly fine with merging the PR as it is for now, and we can address the UI improvements separately afterwards.

Thanks a lot for your patience!

Issue #1023, opened 2026-03-11
Missing enum: 'CHARGING_INTERRUPTED' is not a valid ChargingState


veselytor-wq:
### Please confirm:

- [x] I have searched the existing open and closed issues this problem
- [x] I have updated the integration to the latest available version

### Home Assistant installation and version

Home Assistant Container 2025.11.0

### Car model

Elroq RS

### Integration version

1.30.0

### Bug Description

Describe the bug
When the car is plugged into the charger but the charging process is stopped/paused (e.g., waiting for solar power the next day), the MySkoda API returns the state CHARGING_INTERRUPTED.
The integration crashes and constantly spams the logs because this specific state is missing from the ChargingState enum. This causes a deserialization failure and aggressive retries.

To Reproduce

Plug the cable into the car.

Stop the charging process.

The API returns CHARGING_INTERRUPTED.

The integration fails to deserialize the JSON and throws an error in a loop.

Logs

Python

2026-03-10 12:11:02.149 ERROR (MainThread) [myskoda.rest_api] Failed to deserialize data: {"isVehicleInSavedLocation":false,"status":{"chargePowerInKw":0.0,"state":"CHARGING_INTERRUPTED", ...
ValueError: 'CHARGING_INTERRUPTED' is not a valid ChargingState
mashumaro.exceptions.InvalidFieldValue: Field "state" of type Optional[ChargingState] in ChargingStatus has invalid value 'CHARGING_INTERRUPTED'
Versions

Home Assistant Custom Component MySkoda version: 1.30.0

Underlying myskoda python library version: 2.5.2

### Debug Logs

```bash

```

jeroenribbink:
I also have this message with our Enyaq 60 from 2021

veselytor-wq:
Thank you for the quick reply! @penasp

However, updating the cs.json translation file will unfortunately not fix
the core crash. The error ValueError: 'CHARGING_INTERRUPTED' is not a valid
ChargingState happens deep in the backend during JSON deserialization (via
mashumaro), long before the UI translation layer is even reached.

The state CHARGING_INTERRUPTED needs to be added to the ChargingState Enum
in the underlying myskoda python package first. Adding it only to the HA
translation file won't prevent the loop of crashes and server polling.
Generated by Gemini Pro -- Deep Research

st 11. 3. 2026 v 21:14 odesílatel penasp ***@***.***> napsal:

PR #1024, opened 2026-03-11
Missing enum: 'CHARGING_INTERRUPTED' #1023

penasp:
I fixed only czech, slovak and english files...


penasp:
neni to spravne misto

PR #1047, opened 2026-03-23
chore: update number entities so they use assumed state

WebSpider:
This updates all Numbers entities so they use the assumed_state instead of awaiting an update from the car.

This changes UI behaviour, as it is confusing for users.

**Current behaviour**

- A number entity holds a value
- If it is updated by a user it becomes unavailable while confirming the new value from MySkoda
- The entity becomes available again when an update is received from MySkoda

**New behaviour**

- A number entity holds a value
- If it is updated by a user the update is processed right away and is is marked as an [`assumed_state`](https://developers.home-assistant.io/docs/core/entity/#generic-properties)
- If an update is received from MySkoda, the `assumed_state` mark is removed

Once we get this right for the number entity, all other enties following the same pattern (switch / lock / button) will be updated using the new logic

### DIFF
--- a/custom_components/myskoda/number.py
+++ b/custom_components/myskoda/number.py
@@ -115,7 +115,7 @@ class ChargeLimit(MySkodaNumber):
 
     @Throttle(timedelta(seconds=API_COOLDOWN_IN_SECONDS))
     async def async_set_native_value(self, value: float):  # noqa: D102
-        if self._ensure_not_readonly():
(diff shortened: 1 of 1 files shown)


dvx76:
Happy to try this approach. The current implementation _is_ confusing. Only risk I guess is if an API update happens, still reporting the old value, after the assumed value got set. But not much we can do about that I guess.

PR #1032, opened 2026-03-13
Add dynamic WarningLightSensor for all categories with icons

matisere38:
I retrieved the vehicle's health data.

PR #1069, opened 2026-04-06
Add Warning Light Sensor

matisere38:


WebSpider:
This looks like a duplicate of #1032

matisere38:
Yes, but without error. and base on last official version

WebSpider:
@matisere38 Please fix the tests 👍

@dvx76 this adds one entity for every in-car category, which I think is a bit overkill, given that we already have 50+ entity. 

Care to share your opinion?

matisere38:
Sorry, but I don't understand what I'm supposed to do.
I don't understand the issue

dvx76:
> @dvx76 this adds one entity for every in-car category, which I think is a bit overkill, given that we already have 50+ entity. 
> 
> Care to share your opinion? 

Will take a look this weekend. Away on a trip this week 😊

dvx76:
> @dvx76 this adds one entity for every in-car category, which I think is a bit overkill, given that we already have 50+ entity.
> 
> Care to share your opinion?

@WebSpider I see what you mean. However each category can itself consist of more than one `DefectDetails`. So a single entity would effectively have as attribute(s) a list (or map) of lists. Which is certainly possible, but is tedious to work with in HA (e.g. to extracts the actual warning text and put it on a card in a dashboard).

With the `WarningLightCategory` being a finite set I think it's ok to have separate entities. To keep the standard set of exposed entities under control what we could do is consider having these as `disabled` by default. I've seen other integrations do that as well. Then if a user wants (some of) these they can enable them.

custom_components/myskoda/sensor.py
```diff
@@ -85,6 +86,16 @@ async def async_setup_entry(
         async_add_entities=async_add_entities,
     )
 
+    # --- Dynamic creation of Warning Light sensors ---
+    sensors = []
+    coordinators = hass.data[DOMAIN][config.entry_id][COORDINATORS]
+
+    for vin, coordinator in coordinators.items():
+        for category in WarningLightCategory:
+            sensors.append(WarningLightSensor(coordinator, vin, category))
```
dvx76, line 95: We should only create entities for the warningcategories which are actually exposed for the vehicle  (e.g. combustion engine vehicles have no `ELECTRIC_ENGINE` category).

For example using a `WarningLightSensor.is_supported()` method which checks if the category is present in `vehicle.health.warning_lights`.

custom_components/myskoda/sensor.py
```diff
@@ -993,3 +1004,46 @@ def native_value(self) -> float | None:  # noqa: D102
...
+CATEGORY_ICONS = {
+    "ASSISTANCE": "mdi:car-wrench",
+    "COMFORT": "mdi:seat-recline-normal",
+    "BRAKE": "mdi:car-brake-alert",
+    "ENGINE": "mdi:engine",
+    "ELECTRIC_ENGINE": "mdi:car-electric",
+    "LIGHTING": "mdi:car-light-high",
+    "TIRE": "mdi:tire",
+    "OTHER": "mdi:alert-circle",
+}
```
dvx76, line 1018: Consider using icons which reflect a warning, e.g. `mdi:car-light-alert`, `mdi:car-tire-alert`, ...

custom_components/myskoda/sensor.py
```diff
@@ -993,3 +1004,46 @@ def native_value(self) -> float | None:  # noqa: D102
...
+            entity_category=EntityCategory.DIAGNOSTIC,
+        )
+
+        super().__init__(coordinator, vin)
+
+    @property
+    def native_value(self) -> str | None:
+        """Return defects for this warning category."""
+        if not hasattr(self.vehicle, "health"):
+            return None
```
dvx76, line 1041: nit: Not needed. `self.vehicle.health` will always exist. It may be `None`, but it will always exist.

custom_components/myskoda/sensor.py
```diff
@@ -993,3 +1004,46 @@ def native_value(self) -> float | None:  # noqa: D102
...
+        if not self.vehicle.health:
+            return None
+        for warning in self.vehicle.health.warning_lights:
+            if warning.category == self._category:
+                if warning.defects:
+                    return ", ".join(defect.text for defect in warning.defects)
+                return "OK"
```
dvx76, line 1048: nit: Simplified. Returns empty string if there are no defects, which I think is better than a hardcoded `OK` string.

```suggestion
        if health := self.vehicle.health:
            for warning in health.warning_lights:
                if warning.category == self._category:
                    return ", ".join(defect.text for defect in warning.defects)
```

custom_components/myskoda/sensor.py
```diff
@@ -993,3 +1004,46 @@ def native_value(self) -> float | None:  # noqa: D102
...
+
+    def __init__(self, coordinator, vin, category: WarningLightCategory):
+        self._category = category
+
+        self.entity_description = SensorEntityDescription(
+            key=f"{category.value.lower()}_warning",
+            name=f"{category.value.title()} Warning",
+            translation_key=f"{category.value.lower()}_warning",
+            icon=CATEGORY_ICONS.get(category.value, "mdi:alert-circle"),
+            entity_category=EntityCategory.DIAGNOSTIC,
```
dvx76, line 1032: Let's have these disabled by default.


```suggestion
            entity_category=EntityCategory.DIAGNOSTIC,
            entity_registry_enabled_default=False,
```

custom_components/myskoda/sensor.py
```diff
@@ -993,3 +1004,46 @@ def native_value(self) -> float | None:  # noqa: D102
...
+    def native_value(self) -> str | None:
+        """Return defects for this warning category."""
+        if not hasattr(self.vehicle, "health"):
+            return None
+        if not self.vehicle.health:
+            return None
+        for warning in self.vehicle.health.warning_lights:
+            if warning.category == self._category:
+                if warning.defects:
+                    return ", ".join(defect.text for defect in warning.defects)
```
dvx76, line 1047: Discussion: I don't love having this unpredictable concatenation of (potentially quite long) strings as the actual entity value.

I wonder if it would be better to model these warnings as `BinarySensorEntity` (`binary_sensor.py`), with `device_class=BinarySensorDeviceClass.PROBLEM`, which could be `on` if `defects` is not empty. And then have the list of defect strings in an attribute.

@WebSpider what do you think?

dvx76 (changes requested):
Thanks for contributing to this project!

Main discussion point is around having the concatenated defect texts as entity value.

A few other small things I'd like to see changed before merging this.

matisere38:
Hello, I'm online with all your changes; there's no problem validating them on my end.

Regards

custom_components/myskoda/sensor.py
WebSpider: @dvx76 I absolutely agree on this.

dvx76:
Hi @matisere38 , are you still working on this?

matisere38:
Hi, I'm using the code I shared in my application. As previously discussed, I didn't understand what further steps I needed to take to have it integrated into the official source code. Therefore, I'm no longer posting updates and am keeping my code, which meets my needs.
You can therefore remove the requests and modifications.
Sorry.
Regards

dvx76:
Created #1127 with a reference to this PR in case someone wants to pick it up.

Nornode:
> Therefore, I'm no longer posting updates and am keeping my code, which meets my needs. You can therefore remove the requests and modifications. Sorry. Regards

While I understand your frustration when developers push requirements on to ones code, I think it's sad that you made it all this way without finishing to the end. 
It's okay to ask for help and clarifications again *and again*. 
I have been looking for this data as I had it in the Volvo integration from my previous car and I would have highly appreciated this feature.  

@dvx76   - Any chance you can point @matisere38 in the right direction of the tests required?

matisere38:
Hello, I do indeed make quite a few modifications to scripts I find on GitHub. They're very well-designed, but they don't always meet my needs, as is the case here or with ECOVACS, SMARTTHING, etc. I've been playing around with GitHub, which I'm not very familiar with, and thought it would be interesting to share my modifications, as I was happy to find script templates that I wouldn't know how to develop myself. Unfortunately, I don't have the GitHub code and couldn't answer @dvx76's questions. I'm currently migrating from HOMESEER to HomeAssistant; I've already transferred 80-90% of the devices and events. Once finished, I'll try to better understand GitHub's philosophy, what's expected of me there, and especially to understand compilation errors, knowing that it runs without any problems on my system ;-)
However, feel free to adapt my code modifications if you know how to do them so they're accepted.

Regards

PR #1074, opened 2026-04-14
feat: add new set_departure_timer service

### LINKED ISSUE #392: Feature: Enable pre-conditioning by timer like in App
The official app supports pre-conditioning with a timer. Would be awesome to have similar possibilities like in the main app.

WebSpider:
Fixes #392
 
Finally, the first `action` for this integration, it has been a long time coming :D

This adds a `set_departure_timer` action that lets users fully configure the departure timer including times, type of timer, charging/climatization settings, recurrance, battery target and preferred charging window

Both yaml and json are accepted (since HA automatically converts)

Example action would be:

```yaml
  service: myskoda.set_departure_timer
  data:
    device_id: "abc123def456"
    timer:
      id: 1
      enabled: true
      type: RECURRING
      time: "07:00"
      charging: true
      climatisation: true
      recurringOn: [MONDAY, TUESDAY, WEDNESDAY, THURSDAY, FRIDAY]
      targetBatteryStateOfChargeInPercent: 80
```

Testing still needs to be done, possibly some documentation needs to be added as well, but I figured dropping the code as inspiration doesn't hurt.


README.md
```diff
@@ -84,6 +84,46 @@ This happens independently for each entity.
...
+| `timer.oneOffDay` | No | Single weekday for `ONE_OFF` type |
+| `timer.targetBatteryStateOfChargeInPercent` | No | Target charge level (0–100) |
+
+Example:
+
+```yaml
+action: myskoda.set_departure_timer
+data:
+  device_id: "abc123def456"
+  timer:
```
dvx76, line 111: Can/should we flatten this `timer`? So that each field can get its individual description in the action builder in HA. And should the `device_id` be the _target_ (sorry if this doesn't make sense I don't know much about actions).

E.g. if I set up an action with the tado integration the YAML ends up like this:

```yaml
action: tado.set_climate_timer
target:
  entity_id: climate.bureau
data:
  temperature: 21
  time_period: "04:00:00"
  requested_overlay: MANUAL
```

custom_components/myskoda/__init__.py
```diff
@@ -162,6 +237,15 @@ async def async_setup_entry(hass: HomeAssistant, entry: MySkodaConfigEntry) -> b
     entry.runtime_data = coordinators
 
     await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
+
+    if not hass.services.has_service(DOMAIN, SERVICE_SET_DEPARTURE_TIMER):
+        hass.services.async_register(
+            DOMAIN,
+            SERVICE_SET_DEPARTURE_TIMER,
+            lambda call: _async_handle_set_departure_timer(hass, call),
```
dvx76: From looking at other examples if you just give this the function to call it will automatically be called with the `ServiceCall` object as argument.


```suggestion
            _async_handle_set_departure_timer,
```

Then in that function you can access the `hass` object via `call.hass` (e.g. `dev_reg = dr.async_get(call.hass)`

custom_components/myskoda/__init__.py
```diff
@@ -55,6 +64,72 @@
...
+
+    vin: str | None = next(
+        (ident[1] for ident in device.identifiers if ident[0] == DOMAIN),
+        None,
+    )
+    if vin is None:
+        raise ServiceValidationError(
+            f"Device '{device_id}' VIN not found. Set up the {DOMAIN} integration again"
+        )
+
```
dvx76, line 102: This is totally fine and just FYI in case you prefer this pattern: you can also assume `next` will find something and catch `StopIteration` if it doesn't - then you don't need to default to and check for `None`. E.g.


```suggestion
    try:
        vin = next(ident[1] for ident in device.identifiers if ident[0] == DOMAIN)
    except StopIteration:
        raise ServiceValidationError(
            f"Device '{device_id}' VIN not found. Set up the {DOMAIN} integration again"
        )
```

custom_components/myskoda/services.yaml
```diff
@@ -0,0 +1,45 @@
...
+        recurringOn (list of MONDAY…SUNDAY for RECURRING),
+        targetBatteryStateOfChargeInPercent (0–100),
+        preferredChargingTimes (list of {id, enabled, startTime, endTime}).
+      required: true
+      example:
+        id: 1
+        enabled: true
+        type: RECURRING
+        time: "07:00"
+        charging: true
```
dvx76, line 30: What does `charging` do? I don't see that option when configuring departure timers in the app.

README.md
WebSpider: The tado integration has multiple entities that could be a valid target for the action, we do not, unless we make an entity out of the different timer id's, which seemed counter-intuitive to me. Then again, we also have buttons to enable airco schedules🤔 

The only valid target for us is the vehicle, which is at the device level. 

We could move the device section to `target`, that would make sense.

custom_components/myskoda/services.yaml
WebSpider: I *think* that turning off charging makes it effectively a departure profile, but im not sute tbh 🫣

custom_components/myskoda/__init__.py
```diff
@@ -55,6 +64,72 @@
...
+                vol.Required("id"): vol.In([1, 2, 3]),
+                vol.Required("enabled"): bool,
+            },
+            extra=vol.ALLOW_EXTRA,
+        ),
+    }
+)
+
+
+async def _async_handle_set_departure_timer(
```
dvx76, line 81: This is our first service/action right? Does it have to live within the `__init__.py` or should we put it in a new module? Not sure what the HA convention is.

Also, should executing this be guarded by a check for `coordinator.entry.options.get(CONF_READONLY)` like we have in entities (e.g. `MySkodaButton._ensure_not_readonly()`)?

custom_components/myskoda/__init__.py
```diff
@@ -162,6 +237,15 @@ async def async_setup_entry(hass: HomeAssistant, entry: MySkodaConfigEntry) -> b
     entry.runtime_data = coordinators
 
     await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
+
```
dvx76, line 240: [AI review - I need to look into how services actually work]

> The `has_service` guard means only the first config entry registers it, but `async_unload_entry` never removes it. If the last entry is unloaded, the service lingers and fails at call time with "No coordinator found". Either register/unregister per-entry with refcounting, or use `entry.async_on_unload(...)` to remove when the owning entry unloads

custom_components/myskoda/__init__.py
WebSpider: It is our first action, yes. And i will see on the convention, it would totally make sense to put this in an `actions.py`

I will add the guard for `CONF_READONLY`, thanks, I had forgotten about that mode.

WebSpider: I looked at it, the default scaffolding puts this in `device_action.py`. Will update.

README.md
dvx76: Let's make sure this is streamlined with #1149 

That PR currently defined a YAML schema in line with what I was thinking:

```yaml
action: myskoda.set_charging_profile_time
target:
  device_id: <id of a charging profile device>
data:
  id: 1
  enabled: true
  start_time: "20:00"
  end_time: "06:00"
```

WebSpider: Yeah i already went that road, taking the `renault` integration as an example 👍 

With summer holiday coming, i hope to finish this before my downtime.

PR #1078, opened 2026-04-21
refactor: migrate from hass.data to entry.runtime_data

WebSpider:
Replace `hass.data[DOMAIN][entry_id]` runtime data storage with `entry.runtime_data`, 
flatten the structure of the runtime data (it was a `dict` with only 1 key).

This aligns with `runtime-data` HA quality rule at https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/runtime-data

Tested in an environment running HA 2026.4.4


WebSpider:
@dvx76 i finally got around to testing this, would you be able to review?

custom_components/myskoda/coordinator.py
dvx76: I think the reason we caught `KeyError` here is in case `hass.data` didn't yet contain any of the nested keys. So moving to `self.entry.runtime_data` we may want to also catch `AttributeError` now.

But I'm wondering now why we even need to look up the right coordinator. We're already _in_ the `MySkodaDataUpdateCoordinator` so we can just use `self.myskoda.mqtt`? Right?

```python
if not self.myskoda.mqtt and not self._mqtt_connecting:
    self.entry.async_create_background_task(
        self.hass, self._async_retry_mqtt_connect(), "mqtt"
    )
```

dvx76 (approved):
One comment but LGTM otherwise 👍

PR #1087, opened 2026-05-02
feat: add ActiveVentilationSwitch for ACTIVE_VENTILATION capability

### LINKED ISSUE #1086: Active ventilation switch missing for ICE vehicles without auxiliary heater (ACTIVE_VENTILATION capability)
## Description

Vehicles with the `ACTIVE_VENTILATION` capability (e.g. Skoda Octavia ICE without a parking heater) can start a 10-minute cabin ventilation from the official MySkoda app, but this feature is not exposed in Home Assistant.

## Current behaviour

- The `ACTIVE_VENTILATION` capability is partially handled: `FAN_ONLY` is added to the aux heater climate entity's modes when this capability is present
- However, the aux heater climate entity is **never created** for vehicles that only have `ACTIVE_VENTILATION` (it requires `AUXILIARY_HEATING` or `AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY`)
- Even if the entity existed, `FAN_ONLY` routes to `_start_auxiliary_heating()` instead of the dedicated `start_ventilation()` / `stop_ventilation()` API methods

## Expected behaviour

A switch entity should be available for vehicles with `ACTIVE_VENTILATION` capability, allowing the user to start/stop cabin ventilation — identical to the 10-minute ventilation button in the MySkoda app.

## The API supports it

The `myskoda` library already provides:
- `myskoda.start_ventilation(vin)` → `POST /api/v2/air-conditioning/{vin}/active-ventilation/start`
- `myskoda.stop_ventilation(vin)` → `POST /api/v2/air-conditioning/{vin}/active-ventilation/stop`

## Vehicle

Skoda Octavia (ICE, no auxiliary/parking heater), capability: `ACTIVE_VENTILATION`

tacopiek:
Fixes #1086

## Summary

Adds a `switch` entity for vehicles that support active ventilation (cabin air circulation without heating or cooling) but do not have an auxiliary heater or electric air conditioning — e.g. Skoda Octavia ICE.

This exposes the same "10-minute ventilation" feature available in the official MySkoda app.

## Root cause

The `ACTIVE_VENTILATION` capability was partially handled:
- `FAN_ONLY` was conditionally added to the aux heater climate entity's modes
- But the aux heater entity itself requires `AUXILIARY_HEATING` or `AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY` — so it was **never created** for ACTIVE_VENTILATION-only vehicles
- Even if it had been created, `FAN_ONLY` incorrectly called `_start_auxiliary_heating()` instead of the dedicated ventilation endpoints

## Implementation

New `ActiveVentilationSwitch` in `switch.py`, following the exact same pattern as `WindowHeatingSwitch`:

```python
def required_capabilities(self) -> list[CapabilityId]:
    return [CapabilityId.ACTIVE_VENTILATION]

@property
def is_on(self) -> bool | None:
    if ac := self.vehicle.air_conditioning:
        return ac.state == AirConditioningState.VENTILATION

async def _async_turn_on_off(self, turn_on: bool):
    if turn_on:
        await self._flip_switch(myskoda.start_ventilation(vin))
    else:
        await self._flip_switch(myskoda.stop_ventilation(vin))
```

## Files changed

- `switch.py` — new `ActiveVentilationSwitch` class + registered in `async_setup_entry`
- `translations/en.json` — English label
- `translations/nl.json` — Dutch label

## Test plan

- [ ] Verified on Skoda Octavia (ICE, `ACTIVE_VENTILATION` capability, no aux heater)
- [ ] Switch entity appears after integration restart
- [ ] Turning switch ON activates cabin ventilation (confirmed via MySkoda app)
- [ ] Turning switch OFF stops ventilation
- [ ] `is_on` correctly reflects `AirConditioningState.VENTILATION` from API state
- [ ] Vehicles without `ACTIVE_VENTILATION` capability: switch entity not created

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/custom_components/myskoda/__init__.py
+++ b/custom_components/myskoda/__init__.py
@@ -16,13 +16,14 @@ from myskoda import (
     AuthorizationFailedError,
     MySkoda,
 )
+from myskoda.models.common import Vin
+from myskoda.myskoda import TRACE_CONFIG
 from myskoda.auth.authorization import (
     CSRFError,
     MarketingConsentError,
     TermsAndConditionsError,
     TokenExpiredError,
 )
-from myskoda.myskoda import TRACE_CONFIG
 
 from .const import (
     CONF_FCM_TOKEN,
@@ -30,7 +31,6 @@ from .const import (
     CONF_REFRESH_TOKEN,
     CONF_USERNAME,
     CONF_VINLIST,
-    COORDINATORS,
     DOMAIN,
 )
 from .coordinator import MySkodaConfigEntry, MySkodaDataUpdateCoordinator
@@ -124,7 +124,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: MySkodaConfigEntry) -> b
     async_delete_tnc_issue(hass, entry.entry_id)
     async_delete_spin_issue(hass, entry.entry_id)
 
-    coordinators: dict[str, MySkodaDataUpdateCoordinator] = {}
+    coordinators: dict[Vin, MySkodaDataUpdateCoordinator] = {}
     cached_vins: list = entry.data.get(CONF_VINLIST, [])
 
     try:
@@ -159,8 +159,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: MySkodaConfigEntry) -> b
         await coordinator.async_config_entry_first_refresh()
         coordinators[vin] = coordinator
 
-    hass.data.setdefault(DOMAIN, {})
-    hass.data[DOMAIN][entry.entry_id] = {COORDINATORS: coordinators}
+    entry.runtime_data = coordinators
 
     await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
     entry.async_on_unload(entry.add_update_listener(_async_update_listener))
@@ -171,9 +170,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: MySkodaConfigEntry) -> b
 async def async_unload_entry(hass: HomeAssistant, entry: MySkodaConfigEntry) -> bool:
     """Unload a config entry."""
 
-    coordinators: dict[str, MySkodaDataUpdateCoordinator] = hass.data[DOMAIN][
-        entry.entry_id
-    ].get(COORDINATORS, {})
+    coordinators: dict[Vin, MySkodaDataUpdateCoordinator] = entry.runtime_data
     for coord in coordinators.values():
         if entry.data.get(CONF_REFRESH_TOKEN):
             current_refresh_token = await coord.myskoda.get_refresh_token()
@@ -190,11 +187,7 @@ async def async_unload_entry(hass: HomeAssistant, entry: MySkodaConfigEntry) -> 
             entry_data[CONF_FCM_TOKEN] = coord.myskoda.fcm_token
             hass.config_entries.async_update_entry(entry, data=entry_data)
         await coord.myskoda.disconnect()
-    unload_ok = await hass.config_entries.async_unload_platforms(entry, PLATFORMS)
-    if unload_ok:
-        hass.data[DOMAIN].pop(entry.entry_id)
-
-    return unload_ok
+    return await hass.config_entries.async_unload_platforms(entry, PLATFORMS)
 
 
 async def _async_update_listener(hass: HomeAssistant, entry: MySkodaConfigEntry):
--- a/custom_components/myskoda/binary_sensor.py
+++ b/custom_components/myskoda/binary_sensor.py
@@ -22,7 +22,6 @@ from myskoda.models.info import CapabilityId
 from myskoda.models.status import DoorWindowState, Status
 from myskoda.models.vehicle_connection_status import VehicleConnectionStatus
 
-from .const import COORDINATORS, DOMAIN
 from .coordinator import MySkodaConfigEntry
 from .entity import MySkodaEntity
 from .utils import add_supported_entities
@@ -59,7 +58,7 @@ async def async_setup_entry(
             VehicleInMotion,
             VehicleReachable,
         ],
-        coordinators=hass.data[DOMAIN][config.entry_id][COORDINATORS],
+        coordinators=config.runtime_data,
         async_add_entities=async_add_entities,
     )
 
--- a/custom_components/myskoda/button.py
+++ b/custom_components/myskoda/button.py
@@ -20,7 +20,7 @@ from myskoda.mqtt import OperationFailedError
 
 from aiohttp import ClientResponseError
 
-from .const import API_COOLDOWN_IN_SECONDS, CONF_READONLY, COORDINATORS, DOMAIN
+from .const import API_COOLDOWN_IN_SECONDS, CONF_READONLY, DOMAIN
 from .coordinator import MySkodaConfigEntry, MySkodaDataUpdateCoordinator
 from .entity import MySkodaEntity
 from .utils import add_supported_entities
@@ -37,7 +37,7 @@ async def async_setup_entry(
     """Set up the button platform."""
     add_supported_entities(
         available_entities=[HonkFlash, Flash, WakeUp],
-        coordinators=hass.data[DOMAIN][config.entry_id][COORDINATORS],
+        coordinators=config.runtime_data,
         async_add_entities=async_add_entities,
     )
 
--- a/custom_components/myskoda/climate.py
+++ b/custom_components/myskoda/climate.py
@@ -44,7 +44,6 @@ from .const import (
     API_COOLDOWN_IN_SECONDS,
     CONF_READONLY,
     CONF_SPIN,
-    COORDINATORS,
     DOMAIN,
 )
 from .coordinator import MySkodaConfigEntry, MySkodaDataUpdateCoordinator
@@ -62,7 +61,7 @@ async def async_setup_entry(
 ) -> None:
     add_supported_entities(
         available_entities=[MySkodaClimate, AuxiliaryHeater],
-        coordinators=hass.data[DOMAIN][entry.entry_id][COORDINATORS],
+        coordinators=entry.runtime_data,
         async_add_entities=async_add_entities,
     )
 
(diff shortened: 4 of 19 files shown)


dvx76:
Thanks for your contribution and detailed PR description @tacopiek 

I am wondering if rather than adding a new switch we should update/fix the climate entity to handle this case.

If we _do_ add the switch I think its `forbidden_capabilities()` should return `CapabilityId.AUXILIARY_HEATING` as to avoid adding the switch for `ACTIVE_VENTILATION` vehicles which are already working with the climate entity.

@prvakt do you want to take a look at this?

WebSpider:
I think @dvx76 has a point here. We already have an entity that attempts to handle the case, but doesnt do it quite the way it should in all cases.

Could you update the PR to have the `climate` entity:

- Use the correct set of capabilities 
- Depending on the capabilities present make the right choice in what API endpoint to call

tacopiek:
Thanks for the feedback @dvx76 and @WebSpider!

I've updated the PR as suggested — instead of adding a new switch entity, I've updated `MySkodaClimate` to handle `ACTIVE_VENTILATION` vehicles:

- `is_supported()` now returns `True` for vehicles with `AIR_CONDITIONING` **or** `ACTIVE_VENTILATION`
- For ventilation-only vehicles (ACTIVE_VENTILATION without AIR_CONDITIONING): shows `fan_only` / `off` modes, no temperature control, routes to `start_ventilation` / `stop_ventilation`
- Existing AIR_CONDITIONING behavior is unchanged
- `ActiveVentilationSwitch` has been removed

**Tested on a real ACTIVE_VENTILATION vehicle:**
- Climate entity appears correctly with `fan_only` and `off` modes
- No temperature slider shown
- Turning on starts the ventilation (confirmed the fan physically starts)
- Turning off stops it

**Known limitations observed during testing:**

1. **No timer feedback** — a configured ventilation timer does not appear to have any effect via this entity. The ventilation starts and stops, but there is no indication of how long it will run.

2. **Responsiveness** — the entity is not very snappy. After sending a command, the state may temporarily revert before settling. This appears to be consistent with how `AuxiliaryHeater` works in the same integration: `_operation_in_progress` is cleared as soon as the API call returns, but the Skoda server then sends an MQTT operation event which triggers a coordinator update and clears the optimistic state before the vehicle has actually changed. Combined with the 30-second API throttle, repeated clicks within that window are silently ignored. This seems to be an existing architectural limitation rather than something introduced by this PR.

WebSpider:
Thanks for your update! 

Regarding your known issue 1, the missing remaining time: Would you happen to have a fixture of the ventilation response or related mqtt event? 

We could use that for:

1- Adding tests in the myskoda library
2- See if the information maybe buried in an api response / mqtt message we have not yet seen

tacopiek:
Thanks for your patience @WebSpider and @dvx76! Here are the fixtures and a deeper investigation of the ventilation state behavior.

---

## Fixtures: Skoda Octavia 2022 ICE (ACTIVE_VENTILATION only)

Captured with `myskoda[cli]==2.11.0` directly against the live API.

### Ventilation OFF

<details>
<summary>octavia_ice_ventilation_off.yaml</summary>

```yaml
description: Skoda Octavia ICE - ACTIVE_VENTILATION capability, ventilation off
generation_time: '2026-05-18T17:56:42.524674+00:00'
library_version: 2.11.0
name: octavia_ice_ventilation_off
reports:
- endpoint: air_conditioning
  error: null
  raw: '{"state": "INVALID", "timers": [{"id": 1, "enabled": false, "time": "08:05",
    "type": "ONE_OFF", "selectedDays": ["TUESDAY"]}, {"id": 2, "enabled": false, "time":
    "05:45", "type": "ONE_OFF", "selectedDays": ["TUESDAY"]}, {"id": 3, "enabled":
    false, "time": "07:35", "type": "ONE_OFF", "selectedDays": ["MONDAY"]}], "carCapturedTimestamp":
    "2026-05-18T17:55:45Z", "errors": [{"type": "UNAVAILABLE_VEHICLE_INFORMATION",
    "description": "APIs for obtaining seat heating status are not available"}, {"type":
    "UNAVAILABLE_CHARGING_INFORMATION", "description": "APIs for obtaining power cable
    connection/lock information are not available"}], "outsideTemperature": {"temperatureValue":
    16.0, "temperatureUnit": "CELSIUS", "carCapturedTimestamp": "2026-05-18T16:29:11Z"},
    "runningRequests": [], "steeringWheelPosition": "LEFT", "windowHeatingState":
    {"front": "INVALID", "rear": "INVALID", "unspecified": "INVALID"}}'
  result:
    state: INVALID
    timers:
    - enabled: false
      id: 1
      selected_days: [TUESDAY]
      time: 08:05
      type: ONE_OFF
    - enabled: false
      id: 2
      selected_days: [TUESDAY]
      time: 05:45
      type: ONE_OFF
    - enabled: false
      id: 3
      selected_days: [MONDAY]
      time: 07:35
      type: ONE_OFF
    outside_temperature:
      temperature_value: 16.0
      unit_in_car: CELSIUS
    running_requests: []
    window_heating_state:
      front: INVALID
      rear: INVALID
      unspecified: INVALID
  success: true
  type: get
  url: /v2/air-conditioning/TMOCKAA0AA000000
  vehicle_id: 0
vehicles:
- capabilities:
  - id: ACTIVE_VENTILATION
    statuses: []
  - id: ACCESS
    statuses: []
  - id: MEASUREMENTS
    statuses: []
  - id: STATE
    statuses: []
  # (other capabilities omitted for brevity)
  device_platform: MBB_ODP
  model: Octavia
  model_year: '2022'
  system_model_id: NX54JD
  trim_level: STYLE
```
</details>

### Ventilation ON (physically running, confirmed audibly)

<details>
<summary>octavia_ice_ventilation_on.yaml</summary>

```yaml
description: Skoda Octavia ICE - ACTIVE_VENTILATION capability, ventilation active
generation_time: '2026-05-18T17:57:01.718603+00:00'
library_version: 2.11.0
name: octavia_ice_ventilation_on
reports:
- endpoint: air_conditioning
  error: null
  raw: '{"state": "INVALID", "timers": [{"id": 1, "enabled": false, "time": "08:05",
    "type": "ONE_OFF", "selectedDays": ["TUESDAY"]}, {"id": 2, "enabled": false, "time":
    "05:45", "type": "ONE_OFF", "selectedDays": ["TUESDAY"]}, {"id": 3, "enabled":
    false, "time": "07:35", "type": "ONE_OFF", "selectedDays": ["MONDAY"]}], "carCapturedTimestamp":
    "2026-05-18T17:56:56Z", "errors": [{"type": "UNAVAILABLE_VEHICLE_INFORMATION",
    "description": "APIs for obtaining seat heating status are not available"}, {"type":
    "UNAVAILABLE_CHARGING_INFORMATION", "description": "APIs for obtaining power cable
    connection/lock information are not available"}], "outsideTemperature": {"temperatureValue":
    16.0, "temperatureUnit": "CELSIUS", "carCapturedTimestamp": "2026-05-18T16:29:11Z"},
    "runningRequests": [], "steeringWheelPosition": "LEFT", "windowHeatingState":
    {"front": "INVALID", "rear": "INVALID", "unspecified": "INVALID"}}'
  result:
    state: INVALID
    running_requests: []
    outside_temperature:
      temperature_value: 16.0
      unit_in_car: CELSIUS
  success: true
  type: get
  url: /v2/air-conditioning/TMOCKAA0AA000000
  vehicle_id: 0
```
</details>

---

## Key finding: `/v2/air-conditioning` never reflects ventilation state

Both fixtures — captured ~19 seconds apart, with ventilation physically confirmed running in the second one — return **`state: INVALID`** and **`runningRequests: []`**.

The `VENTILATION` state defined in `AirConditioningState` is never returned by the API for this vehicle. This is not a timing issue: the second fixture was captured after the fan was audibly running.

---

## Impact on the current climate entity implementation

The `hvac_mode` and `hvac_action` properties in `MySkodaClimate` both rely on detecting `AirConditioningState.VENTILATION`:

```python
# climate.py line 234
if ac.state == AirConditioningState.VENTILATION:
    return HVACMode.FAN_ONLY  # never reached for this vehicle

# climate.py line 254
if ac.state == AirConditioningState.VENTILATION:
    return HVACAction.FAN  # never reached for this vehicle
```

Since the API always returns `INVALID`, the entity will always report `off` after the optimistic state clears — even when ventilation is physically running.

**Confirmed via live HA testing:**

| Time     | HA state   | Trigger                        |
|----------|------------|-------------------------------|
| 20:08:04 | `unknown`  | HA startup                    |
| 20:08:32 | `off`      | First coordinator poll (API → INVALID) |
| 20:09:03 | `fan_only` | Clicked in HA UI (optimistic) |
| ~20:09:33| `off`      | Next coordinator update clears optimistic state |

When ventilation is started via the **MySkoda app**, the HA entity **never changes** from `off` — there is no API state update and no MQTT service event that carries the running state.

Note: MQTT CLI subscription was not possible due to Firebase FCM registration requirements outside of HA, so MQTT service events for ventilation could not be directly observed.

---

## Proposed follow-up

1. **For the myskoda library**: These fixtures confirm that `ACTIVE_VENTILATION` vehicles return `state: INVALID` from `/v2/air-conditioning` at all times. A test case using these fixtures would cover capability detection and mode availability, but cannot test running-state detection from the polling API alone.

2. **For the HA integration**: The current implementation is functionally correct for commands (start/stop work), but the state representation is misleading — `off` does not mean ventilation isn't running, it means the API doesn't tell us. A possible improvement would be to return `HVACMode.OFF` only when `state == AirConditioningState.OFF`, and return `None` (unavailable) for `INVALID`, making the uncertainty explicit rather than silently wrong.

Happy to implement suggestion 2 if that direction makes sense to you.

dvx76 (approved):

dvx76:
Thanks for the detailed write-up.

While I understand the logic behavior the suggested additional change, I'm concerned it will cause even more confusion and issues being raised than the state falling back to `HVACMode.OFF` on the next API poll.

Let's merge this as is and see how many issues/questions we get. Then we can still decide to change the behavior, or improve documentation around it.

dvx76 (approved):

PR #1098, opened 2026-05-12
feat: update device tracker behaviour

### LINKED ISSUE #1116: 2026.6.0b0 raising deprecated logs
### Please confirm:

HAOS 2026.6.0b0

### Car model

N.a.

### Integration version

Latest

### Latest working release

N.a.

### Bug Description

Hi,

I’ve updated my HA installation to the first available beta and I get a warning log about this integration using deprecated aliases. Nothing breaking as of now (it will be in a year) and no hurry but just wanted to inform you early. :)

Image

### Debug Logs

```bash
See the screenshot above
```

WebSpider:
We have been asked by the HA core team to align our integration with a HA architecure update, detailed in https://github.com/home-assistant/architecture/discussions/1387

To align with the removal of the `location_name` property in the `device_tracker`, this PR proposes the following changes:

- Remove the `location_name`
~~- Add an attribute called `motion_detected` with a boolean value, indicating if motion is detected~~
~~- Force the `longitude` and `latitude` properties to `None` when the car is in motion~~
- Keep the `latitude` and `longitude`equal to the last API response, unless those are invalid GPS coordinates, including 0/0. They will be set to `None` then, resulting in `Unknown` inside HA. 

This is a breaking change, since we've earlier been telling users to actively use the `location_name` to check if the car is in motion.

Fixes #1116


sonar98:
> 
>     * Force the `longitude` and `latitude` properties to `None` when the car is in motion
> 
When you set lat/lon to None what will happen if you show it on a map? Will it show the last known location? If not maybe keeping the last lat/lon values is a better idea.

WebSpider:
> > ```
> > * Force the `longitude` and `latitude` properties to `None` when the car is in motion
> > ```
> 
> When you set lat/lon to None what will happen if you show it on a map? Will it show the last known location? If not maybe keeping the last lat/lon values is a better idea.

This is one of the interesting questions to answer and part of the reason a discussion on this matter is needed.

My thoughts for setting it to `None` are:

- We know that the vehicle is in motion
- We know that the API does not update when the vehicle is in motion
- We know therefore the data is always incorrect

There are 3 options, the way this integration is set up, since we do not keep historic values in memory at all:

- We show whichever value the API is showing us
- We show a calculated value based on other fields from API responses
- We show a static response

In this PR I chose the latter, so we do not show data that we know is outdated, but this is up for debate. Would it make more sense to show whichever the API is showing us?

sonar98:
> There are 3 options, the way this integration is set up, since we do not keep historic values in memory at all:
> 
> * We show whichever value the API is showing us
> * We show a calculated value based on other fields from API responses
> * We show a static response
> 
> In this PR I chose the latter, so we do not show data that we know is outdated, but this is up for debate. Would it make more sense to show whichever the API is showing us?

I think other integrations only update the lat/lon when there is a new value so the last value is kept until an update comes in.

EDelsman:
I have a hard time figuring out what I would want as location. My gut feeling says not to use none. But to be honest, I do not really know what I think is best.

I use a custom integration Composite device tracker (https://github.com/pnbruckner/ha-composite-tracker) which combines trackers. It basically uses the tracker that last updated its value as the most reliable source when you have multiple trackers. 

I also use a bluetooth tracker in my car to detect home. I would like that to keep working. I have been thinking about trying to figure out who is driving, and use the composite integration to project that persons position on the car when driving.

What is important I think is that it should show it was last seen at the moment when it started driving. Integrations like Google do the same, and they keep the last known positions. I know it makes more sense to assume static position for Google, while for the car we know it is all but that. So in that sense, none would be more accurate.

There would also be the option to use unavailable, But I'm not sure I like that either.

I can live with anything you pick as long as it is clear the location wasn't updated while driving. 

There must remain a way to see if the car is driving, I was used to the external entity for it. So if we have that, we can still be smart in automations.

A final thought: I would like the state to say it is driving if that is at all possible, but i guess that is not allowed anymore? Use the location of a zone with that name? hmmmm. sounds a bit far off too.

What exactly do you mean by calculated position based on other api responses? I think that will likely violate my frozen last seen time wish. It would definitely mess with the composite tracker integration.

WebSpider:
> What is important I think is that it should show it was last seen at the moment when it started driving. Integrations like Google do the same, and they keep the last known positions. I know it makes more sense to assume static position for Google, while for the car we know it is all but that. So in that sense, none would be more accurate.
> 
> There would also be the option to use unavailable, But I'm not sure I like that either.

<s>That is effectively what setting it to `None` does</s>When setting it to `None` it becomes `Unknown` to HA
 
> I can live with anything you pick as long as it is clear the location wasn't updated while driving.
> 
> There must remain a way to see if the car is driving, I was used to the external entity for it. So if we have that, we can still be smart in automations.

Noted. The external entity was removed when going from `skodaconnect` as API to `myskoda` as API, so that has been a while :-)
 
> A final thought: I would like the state to say it is driving if that is at all possible, but i guess that is not allowed anymore? Use the location of a zone with that name? hmmmm. sounds a bit far off too.

We cannot use the `location_name` property of the `device_tracker` anymore in the foreseeable future. You could, in this proposal, build a helper entity based off of the `motion_detected` attribute
 
> What exactly do you mean by calculated position based on other api responses? I think that will likely violate my frozen last seen time wish. It would definitely mess with the composite tracker integration.

Basically it means we can derive a value off of different API responses. I have not looked into it, but maybe the `trip-statistics` endpoint has GPS coordinates hidden somewhere. Anyways, it was mostly there for completeness.

WebSpider:
@dvx76 there seems to be a tendency to keep the gps coordinates reflecting the api output, even when we know the car is moving. 

Do you see any issues in this?

dvx76:
> @dvx76 there seems to be a tendency to keep the gps coordinates reflecting the api output, even when we know the car is moving.
> 
> Do you see any issues in this?

I feel like that goes against the philosophy of Home Assistant. We've had similar discussions in the past for cases where the API becomes unavailable. _Please report the last known SoC instead of Unavailable_. No. Entity updates are either accurate, or unavailable. If you want a _last known value_ of something, that is trivial to setup locally with a helper.

So my preference would be to report Unavailable.

_However_ ... in this case I suppose the argument can be made that we actually _do_ have data (from the API), but we just know the data is outdated (also from the API). So we could "play dumb" and say _well we just update entities with what the API tells us_. I still think we should report unavailable but I'm happy to go the other way if that is your preference.

custom_components/myskoda/device_tracker.py
```diff
@@ -135,13 +136,14 @@ def extra_state_attributes(self) -> dict:
                 "'unmodified_exterior_front' not found, falling back to 'unmodified_exterior_side'."
             )
             attributes["entity_picture"] = renders.get(ViewPoint.EXTERIOR_SIDE)
-        return attributes
-
-    @property
-    def location_name(self) -> str | None:
```
dvx76, line 141: I understand the deprecation period in HA is one year. So perhaps what we can do is already introduce the `motion_detected` attribute, document its use (and remove any references to `location_name`), and announce we'll deprecate it (for e.g. in 6 months)?

I don't think we can detect that it's actually being used, right? (with the intent to log a warning)

WebSpider:
> I feel like that goes against the philosophy of Home Assistant. We've had similar discussions in the past for cases where the API becomes unavailable. _Please report the last known SoC instead of Unavailable_. No. Entity updates are either accurate, or unavailable. If you want a _last known value_ of something, that is trivial to setup locally with a helper.
> 
> So my preference would be to report Unavailable.
> 
> _However_ ... in this case I suppose the argument can be made that we actually _do_ have data (from the API), but we just know the data is outdated (also from the API). So we could "play dumb" and say _well we just update entities with what the API tells us_. I still think we should report unavailable but I'm happy to go the other way if that is your preference.

I did a check this morning and the Android app these days reports the car as "in motion" (specifically: Contact on), and it does update the in-app map, so it must be getting the GPS coordinates from somewhere.

I think my preference here is to just report the last-known value from the API. If that turns out to give `0.00` or something silly during testing, we might have to reconsider.

custom_components/myskoda/device_tracker.py
WebSpider: This is a good idea, I'll prepare a seperate PR for this just and remove the attribute from this one, reserving this for breaking stuff only.

And no, we cannot detect that it is being used, that I know of. The property is there to query from the UI/API, and there is no tracking when someone requests it.

WebSpider:
I overlooked the existance of `binary_sensor.<vehicle>_in_motion`. The extra state attribute doesnt seem needed at all, since it will just duplicate the value of that binary sensor.

EDelsman:
Despite my comments above that I prefer not to have updates while in motion: if the api can update gps coordinates while driving, I'm ok with having the device_tracker reflect what the api says. Users that want the data would otherwise not get access to that data any other way. This is valid behavior if you see the tracker as the direct representation of said api.

If it proves unreliable for my situation, I can always derive my own tracker using the integration and the vehicle in motion sensor.

I would note though, that in the absence of a device_tracker template sensor, it might not be as easy for some people to figure out how to do it. Currently it involves either using the deprecated known_devices.yaml, the device_tracker.see action, the MQTT integration and/or an arbitrary non-device_tracker template entity to get a derived entity.

WebSpider:
I have made some changes:

- We keep the `device_tracker`values to the last GPS coordinates, unless those are 0 or invalid. Then we set it to `None`

dvx76 (approved):
Code LGTM. Need to confirm `vehicle_in_motion` location is deprecated in `README.md` and `entities.md`?

Maybe in the release notes of `v1.33.0` we're about to release we confirm this will be removed in the `v1.34.0`?

WebSpider:
Docs update was done in #1101, which will be part of `1.33.0` release.

dvx76:
> Docs update was done in #1101, which will be part of `1.33.0` release.

Right, but when this PR merges it will no longer _be removed in a future release_, it will actually be gone. So the suggestion was to completely remove all mentions of `vehicle_in_motion` in this PR (or maybe something like _removed in v1.34.0_)?

dvx76 (approved):
🚀

PR #1102, opened 2026-05-20
refactor: split into multiple coordinators to reduce API usage

### LINKED ISSUE #1056: Todo: Reimplement fetching data
With the amount of endpoints increasing over time, and the API credits staying the same, users with a very low poll interval start to run into trouble (see #1053).

These users currently give us the early warning that we need to look at our data fetching. Currently we fetch the full vehicle regardless, and we should start thinking about splitting the code:

- Implement multiple `DataUpdateCoordinator`s, ideally one per <s>endpoint</s> update interval
- Find a way to map the entities to it's `DataUpdateCoordinator`
- Find a way to distinguish between sensors that need frequent and less frequent updates
- Find a mapping between events and related entities / coordinators

This will give us a more robust structure going forward, as the features, endpoints and eventtypes will only increase as Skoda evolves its API.

WebSpider:
Fixes #1056

Previously a single coordinator fetched all endpoints on every poll. Fast-changing data (charging, status, AC, positions) is now handled by the primary coordinator at the user-configured interval; slow-changing data (trip statistics, maintenance, health, vehicle info) is handled by a new `MySkodaSlowCoordinator` on a fixed interval.

`_on_mqtt_event` no longer calls `async_set_updated_data` unconditionally: entity updates on MQTT now only fire for `OperationEvents`. This prevents unneeded load on the HomeAssistant State Engine

All entities declare their coordinator domain via coordinator_domain (default "primary"); slow-domain sensors override this to "slow".

TODO: Rigorous testing needed


dvx76:
@WebSpider before I start reviewing this, want to check if you will be able to continue this PR. Otherwise I might pick this up at some point.

WebSpider:
I started this, i plan to finish it 👍

Issue #1105, opened 2026-05-21
odo meter, km until next oil service, etc. not updating anymore


n300home:
### Please confirm:

- [x] I have searched the existing open and closed issues this problem
- [x] I have updated the integration to the latest available version

### Home Assistant installation and version

HAOS 2026.5.3

### Car model

Superb 3 iV

### Integration version

1.32.2

### Latest working release

N/A

### Bug Description

Hi there,
Since around 8th of April all my mileage sensors seams to be frozen. Although they are available, they don't get any updates anymore.

Any Ideas?

BR
Manuel

<img width="2032" height="511" alt="Image" src="https://github.com/user-attachments/assets/aaac2a91-3f8b-4d1a-b7ff-55999b06b129" />

[config_entry-myskoda-01JP7PD92W67M7BK4B67YSD9MC.json](https://github.com/user-attachments/files/28108995/config_entry-myskoda-01JP7PD92W67M7BK4B67YSD9MC.json)

### Debug Logs

```bash

```

WebSpider:
Interesting issue! 

This is in your debug logs:

```json
"maintenance_report": {
                  "timestamp": "2026-05-21T15:09:25.794760+00:00",
                  "captured_at": "2079-07-05T17:14:00+00:00",
                  "mileage_in_km": 44433,
                  "inspection_due_in_days": 449,
                  "inspection_due_in_km": 23200,
                  "oil_service_due_in_days": 449,
                  "oil_service_due_in_km": 23200
```

`captured_at` is way in the future! 

This would cause our caching to never update, we need to fix that.

n300home:
Haa! of course. My Superb had a broken OCU, which was replaced some weeks ago. One of the failure was, that it interpreted the date time from GPS completely wrong. The time was correct, but the date was really something with 2079. Which endet up with connection issues to Skoda cloud and navigation issues. Also Android Auto wasn't able to connect, due to TLS Handshake issues. What is clear, when the device thinks all certificates are expired since decades :D

Is the date from   "captured_at": "2079-07-05T17:14:00+00:00" still received from Cloud API, or is it a relict from the time my OCU was defective?

WebSpider:
I had a good look at the code, and there are several points where this code handles the "broke vehicle timestamp" incorrect, but unfortunately, the `maintenance` endpoint results are not cached. We are getting this timestamp directly from MySkoda.

I also suspect the MySkoda app shows incorrect mileage, etc

[...]

PR #1109, opened 2026-05-26
fix: skip image cache flush when car_captured_timestamp is in the future

WebSpider:
When a vehicle reports a timestamp far in the future, the image cache timestamp never changes.

Skip the cache flush when car_captured_timestamp exceeds the allowed clock skew tolerance (4h), and emit a warning.

Relates to #1105


dvx76 (approved):
LGTM!

PR #1123, opened 2026-06-05
fix: remove redundant fcm token refresh

dvx76:
This never worked because `MySkoda.enable_mqtt()` would never raise or return while retrying/reconnecting.

https://github.com/skodaconnect/myskoda/pull/592 moves the logic to refresh the FCM tokem into the MySkodaMQTTClient and ensures it will eventually raise an exception when continuously failing to connect.

---

Tested in HA, both the success path where the MQTT client refreshes the FCM token and the failure path where it raises an exception eventually and the coordinator schedules another attempt in 300s.


WebSpider (approved):
Lgtm

PR #1141, opened 2026-06-30
fix: add position fallback to motion-detection binary sensor

### LINKED ISSUE #1132: New Binary Sensor vehicle_in_motion not existing
### Please confirm:

HAOS 2026.06

### Car model

Octavia 2018

### Integration version

1.34.0

### Latest working release

_No response_

### Bug Description

Good Morning,

First of all, thanks for this great integration 👍😊.

I am not sure if this is a bug or if the problem is sitting in front of the monitor 😅.
After updating to Version 1.34.0 I am not able to find the new binary sensor for vehicle_in_motion. Do you have any idea what I am doing wrong? I can't find any entries in the log concerning this.

Regards,
Chris

### Debug Logs

```bash

```

WebSpider:
This adds the logic for the old `device_tracker`'s movement detection to the new `binary_sensor` as a fallback and creates the sensor whenever the old `device_tracker` was able to detect movement.

Fixes #1132

### DIFF
--- a/custom_components/myskoda/binary_sensor.py
+++ b/custom_components/myskoda/binary_sensor.py
@@ -436,7 +436,7 @@ class VehicleInMotion(VehicleConnectionBinarySensor):
     )
 
     def _positions(self) -> Positions | None:
-        if vehicle_positions := self.vehicle.positions
+        if vehicle_positions := self.vehicle.positions:
             return vehicle_positions
 
     def _motion_detection(self) -> Error | None:
--- a/custom_components/myskoda/binary_sensor.py
+++ b/custom_components/myskoda/binary_sensor.py
@@ -20,6 +20,12 @@ from myskoda.models.common import (
 )
 from myskoda.models.info import CapabilityId
 from myskoda.models.status import DoorWindowState, Status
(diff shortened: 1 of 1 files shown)


dvx76:
I should probably start doing reviews _before_ making changes and opening PRs 🤦 (#1142)

dvx76 (approved):
At least we had the same approach 🙈

custom_components/myskoda/binary_sensor.py
```diff
@@ -435,10 +441,28 @@ class VehicleInMotion(VehicleConnectionBinarySensor):
...
+                return next(
+                    err for err in pos.errors if err.type == ErrorType.VEHICLE_IN_MOTION
+                )
+
     @property
     def is_on(self) -> bool | None:
         if cs := self._connection_status():
             return cs.in_motion
+        if motion_error := self._motion_detection():
+            return motion_error.type == ErrorType.VEHICLE_IN_MOTION
```
dvx76, line 460: If `self.vehicle.positions` is present and there are no errors we probably want to return `False` instead of `None`.

PR #1145, opened 2026-07-06
Remove battery_level property from device_tracker

roofburner:
As per https://developers.home-assistant.io/blog/2026/06/15/device-tracker-changes/:

The battery_level property has been deprecated in all device tracker base classes, and will stop working in Home Assistant Core 2027.7. Integrations should communicate battery level via a battery sensor instead.

mySkoda integration already uses seperate sensor (sensor.skoda_enyaq_battery_percentage)


dvx76 (approved):
LGTM! Thanks.

Issue #685, opened 2023-04-11
Charging locations


are0002:
This is a feature request to monitor when the car is in a configured charge location. 

You can configure in the car a location to set up some charging options like: departure time, preferred charging hours, etc. Once you car is in a configured location, the mobile app detects and let you choose to charge immediately or use the charging location configuration.

It will be great to have the option to monitor:
1 - If the car is in a configured location.
2 - If the selected option is charge using location configuration or immediately.

Thanks in advanced.

Regards,

ev0lister:
+1 - bumping this issue as I also have a use-case for this

chr1s1:
is there any update from the API side? This would ease a lot of automation issues on my end, where I always want to plug in my car and only charge it when the PV produces enough engery.

WebSpider:
We have been kind of blocked by how to handle this nicely in HomeAssistant, since the HA UI does not support proper rendering of list of dictionaries.

It looks like we will implement this in a way that requires you to fill out the complete dictionary in a service, and we'll be going forward with this as time allows. Summer holiday usually means either a lot of coding, or a lot of partying, so YMMV 💃

WebSpider:
Well, I didn't get much coding done 🤣

bassduh:
Is this also supposed add the ability to get/set the location based charger settings? In example charge limit and reduced AC charging current (approx 5,5kW instead of 11kW).

WebSpider:
> Is this also supposed add the ability to get/set the location based charger settings? In example charge limit and reduced AC charging current (approx 5,5kW instead of 11kW).

Yeah, thats the `maxChargingCurrent`

WebSpider:
> Is this also supposed add the ability to get/set the location based charger settings? In example charge limit and reduced AC charging current (approx 5,5kW instead of 11kW).

Just as an FYI: It is almost always better to use the wallbox to limit charging currents. Offcourse, when charging in the public space, you usually don't have this type of control over the wallbox.

bassduh:
> > Is this also supposed add the ability to get/set the location based charger settings? In example charge limit and reduced AC charging current (approx 5,5kW instead of 11kW).
> 
> Just as an FYI: It is almost always better to use the wallbox to limit charging currents. Offcourse, when charging in the public space, you usually don't have this type of control over the wallbox.

I do agree on that - whenever you have the option.
In my particular situation, I unfortunately don't have the option to change settings myself 🤨

f1wade:
[...]

PR #1149, opened 2026-07-12
feat: add charging profile support and action to change preferred charging times

max-2468:
Hello all,

based on the support to edit a charging time in a charging profile in the MySkoda library I have added support for both charging profiles as well as an action to change these to the HA integration.

As there may be multiple charging profile the integration adds a new device per charging profile. The information for each profile is shown as multiple sensors for each device. These are updated based on regular polling. One sensor shows whether the vehicle is currently at the given location, another shows detailed information about preferred charging times as well as charging timers.

The action `myskoda.set_charging_profile_time` allows changing a preferred charging time using the api added to the MySkoda library.

This addresses issue #685.

Test: installed on a local HA instance with a productive MySkoda Account
- multiple charging profiles are shown
- changing the preferred charging times works on multiple profiles
- error messages when providing invalid data to the action work as expected (executing on invalid device, providing an invalid id)

Not tested:
- behaviour with multiple vehicles
- behaviour when vehicle does not support charging profiles. There's a guard on the CHARGING_PROFILES Capability so it should work.

Code was AI generated

Regards,
max


custom_components/myskoda/coordinator.py
```diff
@@ -121,6 +124,25 @@ def __init__(
         self._mqtt_retry_scheduled: bool = False
         self._startup_called: bool = False
 
+    async def _fetch_charging_profiles(
```
dvx76: issue(blocking): This should be handled in the MySkoda library, as part of normal data fetch/refresh, instead of requiring the client (here the integration coordinator) to separately deal with it.

`ChargingProfiles` becomes a new field in `Vehicle`.

README.md
```diff
@@ -72,6 +72,26 @@ This integration interacts with the Skoda API in the same way the official MySko
...
 
+### Actions
+
+#### `myskoda.set_charging_profile_time`
+Updates one preferred charging time window of a charging profile (location, e.g. "Home"). Each configured charging profile is exposed as its own device, nested under the vehicle device; target that device with `device_id`.
+
+The `id` must match an existing time window on that profile (visible as an attribute on its `Preferred Charging Times` sensor) — you cannot create new time windows this way, only update existing ones.
+
+```yaml
+action: myskoda.set_charging_profile_time
```
dvx76: Let's make sure this is streamlined with #1074

README.md
```diff
@@ -72,6 +72,26 @@ This integration interacts with the Skoda API in the same way the official MySko
 
 Refer to [docs/entities.md](docs/entities.md) for the full list of all entities created by the integration.
 
+### Actions
+
+#### `myskoda.set_charging_profile_time`
```
dvx76: suggestion: rename to just `set_charging_profile` since this could also be used to only disable a profile?

custom_components/myskoda/device_action.py
```diff
@@ -0,0 +1,130 @@
...
+
+        for profile in profiles.charging_profiles:
+            expected_identifier = (DOMAIN, f"{vin}_charging_profile_{profile.id}")
+            if expected_identifier in device.identifiers:
+                return coordinator, profile
+
+    return None
+
+
+async def _async_handle_set_charging_profile_time(call: ServiceCall) -> None:
```
dvx76: issue(blocking): At the end of this function we should make sure data is refreshed so the updated charging profiles are immediatelly reflected in HA.

#1074 simply calls `await coordinator.async_request_refresh()`, which I _think_ is ok.

custom_components/myskoda/sensor.py
```diff
@@ -85,6 +90,17 @@ async def async_setup_entry(
         coordinators=config.runtime_data,
         async_add_entities=async_add_entities,
     )
+    add_supported_charging_profile_entities(
+        available_entities=[
+            ChargingProfileTargetStateOfCharge,
+            ChargingProfileMinimumStateOfCharge,
+            ChargingProfileMaxChargingCurrent,
+            ChargingProfileAutoUnlockPlug,
+            ChargingProfileChargingTimes,
```
dvx76: question: do we really need separate entities for every single field of a charging profile? What happens when someone has 5 profiles? They end up with 30 additional entities?

Compare with for example how AC Timers are implemented (https://github.com/skodaconnect/homeassistant-myskoda/blob/e121c20ac76028e5246701241531cf554cfe0f6e/custom_components/myskoda/switch.py#L668). They are just one binary switch (per timer) which can enabled/disable the timer. All other settings (type, time, days, ...) are exposed as additional attributes on that same switch entity.

Charging profiles could be implemented the same way. With binary switches which would allow to directly enable/disable a profile (no action required). The action is then still used to update/set any of the other parameters.

dvx76 (changes requested):
Many thanks for contributing to this project! 🙏 

These are some initial comments after a first quick pass over the proposed changes.

custom_components/myskoda/sensor.py
max-2468: that is certainly possible. Although then the charging timers might have to be represented as switches as well, even though they cannot be changed yet. What is your opinion? Regarding the total number of entities: with 4 or 7 entities per charging profile the number will not change that much. Somebody with 5 charging profiles might just have to live with all those entities - probably an extreme use case though.

README.md
max-2468: It can be used to enable/disable a specific preferred charging time within a charging profile. The other data in a charging profile cannot be changed yet, although changing the MySkoda api to allow this should be relatively easy. For the current functionality, renaming makes sense. How about 'set_preferred_charging_times'?

custom_components/myskoda/coordinator.py
max-2468: Adressed by https://github.com/skodaconnect/myskoda/pull/621, can be refactored here afterwards.

README.md
dvx76: Sounds good

custom_components/myskoda/sensor.py
dvx76: I think I need to play around with charging profiles a bit to better understand the functionality. I've never used it. Will try to do that this week before going on vacation 😁

README.md
max-2468: Action renamed to `set_preferred_charging_time`

custom_components/myskoda/sensor.py
max-2468: I have added switches for all ChargingTimers as well as the Charging Times of a profile. The structure of entities below a Charging Profile device now pretty much matches what is shown in the app, which should therefore be unsurprising to a user:

- Charging Timer 1-3 (Switch, not toggleable at the moment)
- Preferred Charging 1-4 (Switch, toggleable, will update the server)
- Other attributes given for the profile (not editable): Minimum SoC, Target SoC, Max/Reduced Charging Current, Aut-unlock Plug

Comparing the API with the MySkoda app it's pretty clear that the app updates all information of a charging profile at once when saving, so that multiple changes are possible at once. Individually making changes and immediately updating it to the servers might result in a lot of calls. At the moment the MySkoda api is designed to update a single preferred charging time (see below), but a more general approach might be useful if all data should be editable from Home Assistant.

My use case for this change is setting the preferred charging time according to EPEX energy prices, so that the car will automatically start charging when energy is cheapest without having to think about plugging/unplugging it while being parked in an underground garage. This already works using an additional HA component on my local machine. Now I just need a wall box there...

dvx76: Thanks for the updates. I don't think I'm going to get around to play with this before I go on vacation so I'm just going to focus on the code and then get this merged and released for you.

custom_components/myskoda/device_action.py
dvx76: This is still outstanding. Let me know if you think we shouldn't do this.

dvx76:
Can you rebase? You have some merge conflicts to be addressed.

custom_components/myskoda/device_action.py
max-2468: I am not sure. Doing some tests to change a profile from both the mySkoda app and Home Assistant show that the changes get propagated to Home Assistant via MQTT trigger and the following refresh. I think (but that needs to be checked in more detail) that even the Myskoda app only shows the updated values after they were successfully transferred to the car (car is parking in an underground garage with not great connectivity => change profile in app & save => app shows "activating car" => the edited charging profile still shows the old value and a disabled "save" button => updates in the app after the car comes online)

Only updating the state of the entities after they were successfully transferred to the car makes sense in my opinion, but feedback to the user by somehow disabling the entities in HA might make sense until MQTT reports the operation successful.

dvx76: Ok fair enough. Can always be added later.

custom_components/myskoda/switch.py
```diff
@@ -784,3 +796,103 @@ class ACTimer3(ACTimerSwitch):
...
+            }
+        return {}
+
+    @Throttle(timedelta(seconds=API_COOLDOWN_IN_SECONDS))
+    async def _async_turn_on_off(self, turn_on: bool):
+        """Internal method to have a central location for the Throttle."""
+        myskoda = self.coordinator.myskoda
+        action = "on" if turn_on else "off"
+        if times := self.charging_time:
+            times.enabled = turn_on
```
dvx76: This will mutate the current `charging_time` object in the coordinator. This can result in the switch flipping on and off in the UI after an API poll. It will also incorrectly leave the switch flipped on if there is an error/exception during the `set_preferred_charging_times` API call.

Instead consider a clean new `ChargingTimes` object to pass into the `set_preferred_charging_times` call. E.g.

```python
    payload = ChargingTimes(
        id=times.id,
        enabled=turn_on,
        start_time=times.start_time,
        end_time=times.end_time,
    )
```

custom_components/myskoda/translations/en.json
```diff
@@ -270,6 +273,30 @@
...
+            },
+            "charging_profile_auto_unlock_plug": {
+                "name": "Unlock Plug when Charged",
+                "state": {
+                    "permanent": "Permanent",
+                    "on": "On",
+                    "off": "Off"
+                }
+            },
+            "charging_profile_charging_times": {
```
dvx76: This translation key is no longer used (same in `en.json`)

README.md
```diff
@@ -72,6 +72,26 @@ This integration interacts with the Skoda API in the same way the official MySko
 
 Refer to [docs/entities.md](docs/entities.md) for the full list of all entities created by the integration.
 
+### Actions
+
+#### `myskoda.set_preferred_charging_time`
+Updates one preferred charging time window of a charging profile (location, e.g. "Home"). Each configured charging profile is exposed as its own device, nested under the vehicle device; target that device with `device_id`.
+
+The `id` must match an existing time window on that profile (visible as an attribute on its `Preferred Charging Times` sensor) — you cannot create new time windows this way, only update existing ones.
```
dvx76: I think the attributes are now on the switch and the _Preferred Charging Times_ no longer exists.

custom_components/myskoda/sensor.py
```diff
@@ -1017,3 +1032,76 @@ def native_value(self) -> float | None:  # noqa: D102
...
+    """Base class for sensors representing a single charging profile (location)."""
+
+
+class ChargingProfileTargetStateOfCharge(MySkodaChargingProfileSensor):
+    """Target state of charge configured for this charging profile."""
+
+    entity_description = SensorEntityDescription(
+        key="charging_profile_target_state_of_charge",
+        translation_key="charging_profile_target_state_of_charge",
+        device_class=SensorDeviceClass.BATTERY,
```
dvx76: Not sure this should be a `SensorDeviceClass.BATTERY` since it's a target %, not an actual battery % (e.g. the existing `TargetBatteryPercentage` doesn't do this either)

dvx76 (changes requested):

custom_components/myskoda/sensor.py
max-2468: fixed

README.md
max-2468: yes, the attributes are on the switches, those are "Preferred charging times" switches. changed sensor => switch.

custom_components/myskoda/translations/en.json
max-2468: removed

dvx76 (approved):
Thanks for these last updates!

PR #1151, opened 2026-07-12
feat(climate): add camping mode

tamcore:
This exposes camping mode as a "camping" preset on the air-conditioning climate entity, gated on the CAMPING_MODE capability.

testing done:
- with HVAC off; set preset=camping; start_camping with the current AC target temperature called and confirmed by Myskoda app right away
- with HVAC on and preset=camping; selecting preset=none calls stop_camping and turns everything off
- with HVAC on and preset=camping; turning off AC, conditionally calls stop_camping (as calling stop_ac would not work)
- regular AC on/off still works

It also adds a timestamp sensor for campingMode.endsAt, so the remaining time is visible in HA. I feel like it partially duplicates the "AC Time to reach target temperature" entity. But putting both AC-related time sensors into this entity, would require renaming.

Depends on https://github.com/skodaconnect/myskoda/releases/tag/v2.15.0, so bumped pyproject.toml / uv.lock accordingly.

### DIFF
--- a/custom_components/myskoda/climate.py
+++ b/custom_components/myskoda/climate.py
@@ -10,6 +10,7 @@ from homeassistant.components.climate import (
     ClimateEntityDescription,
 )
 from homeassistant.components.climate.const import (
+    PRESET_NONE,
     ClimateEntityFeature,
     HVACAction,
     HVACMode,
@@ -52,6 +53,8 @@ from .utils import add_supported_entities
 
 _LOGGER = logging.getLogger(__name__)
 
+PRESET_CAMPING = "camping"
+
 
 async def async_setup_entry(
     hass: HomeAssistant,
@@ -72,6 +75,7 @@ class OptimisticAttribute(StrEnum):
     TARGET_TEMPERATURE = "target_temperature"
     HVAC_MODE = "hvac_mode"
     HVAC_ACTION = "hvac_action"
+    PRESET_MODE = "preset_mode"
 
 
 class MySkodaClimateEntity(MySkodaEntity, ClimateEntity):
@@ -168,6 +172,24 @@ class MySkodaClimateEntity(MySkodaEntity, ClimateEntity):
         finally:
             self._operation_in_progress = False
 
+    async def _start_camping(self, temperature: float) -> None:
+        self._ensure_not_readonly()
+        self._operation_in_progress = True
+        try:
+            await self.coordinator.myskoda.start_camping(
+                self.vehicle.info.vin, temperature
+            )
+        finally:
+            self._operation_in_progress = False
+
+    async def _stop_camping(self) -> None:
+        self._ensure_not_readonly()
+        self._operation_in_progress = True
+        try:
+            await self.coordinator.myskoda.stop_camping(self.vehicle.info.vin)
+        finally:
+            self._operation_in_progress = False
+
     async def _start_ventilation(self) -> None:
         self._ensure_not_readonly()
         self._operation_in_progress = True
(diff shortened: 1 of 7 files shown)


custom_components/myskoda/climate.py
```diff
@@ -256,6 +298,22 @@ def hvac_action(self) -> HVACAction | None:  # noqa: D102
...
+    @property
+    def preset_modes(self) -> list[str] | None:  # noqa: D102
+        if self._supports_camping():
+            return [PRESET_NONE, PRESET_CAMPING]
+        return None
+
+    @property
+    def preset_mode(self) -> str | None:  # noqa: D102
+        if not self._supports_camping():
+            return None
```
dvx76, line 310: nit: I think this is redundant and we'll fall through to the final return is camping mode isn't supported.

custom_components/myskoda/climate.py
```diff
@@ -311,11 +369,23 @@ async def async_set_hvac_mode(self, hvac_mode: HVACMode):  # noqa: D102
             except (ClientResponseError, OperationFailedError) as exc:
                 self._unset_optimistic_data(OptimisticAttribute.HVAC_MODE)
                 _LOGGER.error("Failed to start air conditioning: %s", exc)
+        elif self._supports_camping() and self._is_camping_active():
```
dvx76, line 372: note to self

edit: lol this was a placeholder in my review I forgot to update 🙈 

meant to say this function is turning in a bit of an if-condition spaghetti and is ripe for a refactor. But of of scope so need to create an issue work on it.

custom_components/myskoda/climate.py
```diff
@@ -311,11 +369,23 @@ async def async_set_hvac_mode(self, hvac_mode: HVACMode):  # noqa: D102
...
+            # Camping keeps the AC running; a plain _stop_air_conditioning would not
+            # end it, so call _stop_camping instead.
+            self._set_optimistic_data(OptimisticAttribute.PRESET_MODE, PRESET_NONE)
```
dvx76, line 375: do after?

custom_components/myskoda/climate.py
```diff
@@ -328,6 +398,37 @@ async def async_turn_on(self):  # noqa: D102
     async def async_turn_off(self):  # noqa: D102
         await self.async_set_hvac_mode(HVACMode.OFF)
 
+    @Throttle(timedelta(seconds=API_COOLDOWN_IN_SECONDS))
+    async def async_set_preset_mode(self, preset_mode: str) -> None:  # noqa: D102
+        if not self._supports_camping():
+            return
```
dvx76, line 404: suggestion/question: this "shouldn't happen" right? Should we instead have an `else` below logs warning like `"{preset_mode} is not supported"`?

custom_components/myskoda/climate.py
```diff
@@ -328,6 +398,37 @@ async def async_turn_on(self):  # noqa: D102
...
+
+        if preset_mode == PRESET_CAMPING:
+            if not (ac := self._air_conditioning()):
+                _LOGGER.debug(
+                    "Cannot start camping mode: air conditioning unavailable."
+                )
+                return
```
dvx76: nit: I think this can't happen with the preset only being available when camping mode is available? So might as well leave the extra check out here.

custom_components/myskoda/climate.py
```diff
@@ -328,6 +398,37 @@ async def async_turn_on(self):  # noqa: D102
...
+            if not (target_temperature := ac.target_temperature):
+                _LOGGER.debug("Cannot start camping mode: no target temperature set.")
+                return
```
dvx76: question: Can this happen and if so, is there a way to get more direct feedback to the user (I think you can raise `ServiceValidationError` for that? Or at least raise the log severity to warning as many users don't have debug enabled by default.

custom_components/myskoda/climate.py
```diff
@@ -328,6 +398,37 @@ async def async_turn_on(self):  # noqa: D102
...
+            self._set_optimistic_data(OptimisticAttribute.PRESET_MODE, preset_mode)
```
dvx76, line 414: issue: I think we also want to optimistically set/unset `HVAC_MODE` (to `HVACMode.HEAT_COOL/HVACMode.OFF`) so that the climate entity immediatally shows AC as on (/off), which it will be, when turning camping mode on (/off).

custom_components/myskoda/sensor.py
```diff
@@ -187,6 +188,25 @@ def extra_state_attributes(self) -> dict[str, Any]:
         return attributes
 
 
+class CampingModeEndsAt(MySkodaSensor):
+    """Report when camping mode will automatically end."""
+
+    entity_description = SensorEntityDescription(
+        key="camping_mode_ends_at",
```
dvx76, line 195: issue: Can you add an icon for this entity? Maybe `mdi:tent`?

dvx76 (changes requested):
Very nice! Thanks for your work on this 💯 .

A few issues and questions, but nothing major.

custom_components/myskoda/climate.py
tamcore: While you're functionally correct and it should not make a user-facing difference, keeping it seems more correct to me. Because, if CAMPING_MODE is not supported at all, we return None (= no presets, no nothing). Otherwise, we would fall through to PRESET_NONE (= could imply support for presets)

Again, either way makes no difference for the user. If you prefer me to remove this bit, i'll do it.

custom_components/myskoda/sensor.py
tamcore: Done.

custom_components/myskoda/climate.py
tamcore: `else` warning implemented at the end down bellow

custom_components/myskoda/climate.py
tamcore: Added a `ServiceValidationError`. I guess the community is not too uniform on this. For example, the Kia/Hyundai integration just takes it as is, Tesla just returns silent and ha-porscheconnect hardcodes 20°C.

custom_components/myskoda/climate.py
tamcore: Implemented. Didn't think about it before.

custom_components/myskoda/climate.py
tamcore: In line 356 HVAC mode is also set directly. So i'd argue to keep this. If the await below errors out, the except resets both anyways.

dvx76: Sorry this was also a placeholder I meant to refine before posting :) 

But you got what I meant. Fair enough though, let's stay consistent.

dvx76 (approved):
Good stuff 🚀

